PR #6, opened 2025-07-01
Multi-project import

ftrotter-gov:
Not actually sure this pull request should be merged. 

It might be better just to have these in separate branch for now...


ftrotter-gov:
Hey Sarah, I kept a json key from serpapi that allowed you to find the "next page" of search results. Whch apparently had embedded credentials in the GET url. 

I have removed the files from HEAD but in order to fix it permenantly we will need to re-upload a fresh copy of the repo without the history.. Or use git-filter-repo to drill out the history. Either way I suspect that we need prvileges to fix the problem. 

I have revoked my serp API keys, so there is no danger anymore, but the gitleaks action will fail until we do this.

PR #8, opened 2025-07-11
Sjp/gitleaksignore

spopelka-dsac:
This commit adds all previously failing secret fingerprints to a .gitleaksignore file (a la https://github.com/gitleaks/gitleaks/blob/master/.gitleaksignore) so that gitleaks will stop yelling at us for future commits. Woo!


spopelka-dsac:
Look at that sweet, sweet passing secrets check! image

PR #9, opened 2025-07-11
SQL standards

ftrotter-gov:
All of the ideas from Sarah's preferred SQL standard are encorporated as simple single sentence rules. Additional rules are added because I still have ambitions to auto-generate data editing forms.

I also wanted to enforce as many of the rules as possible using sqlfluff and so there is a .sqlfluff config that does that were possible. 

For future work, there are a handful of the rules that are not currently enforceable with sqlfluff that could be in they were made into custom rules... and the rest could be enforced by an LLM in a github action at pull request time.


subproject/tinman_SQL_schema_standard/dot.sqlfluff
spopelka-dsac, line 1: Why is this necessary?

subproject/tinman_SQL_schema_standard/.gitignore
spopelka-dsac, line 1: A lot of this is likely duplicative of the root `.gitignore` file

subproject/tinman_SQL_schema_standard/README.md
```diff
@@ -0,0 +1,77 @@
+TinMan SQL Naming Conventions
+=========
+
+Gievs structure to databases with heart.
```
spopelka-dsac, line 4: "Gives"

subproject/tinman_SQL_schema_standard/README.md
```diff
@@ -0,0 +1,77 @@
...
+* [SQLFluff default rules](https://docs.sqlfluff.com/en/stable/configuration/default_configuration.html)
+
+We intend to write additional sqlfluff rules to support the rules that sqlfluff does not already support when possible. eventually. given infinite time and resources.
+
+⸻
+
+Rules Not Currently Enforceable by SQLFluff
+------
+
+These require either a custom rule, external linter, or manual review:
```
spopelka-dsac, line 19: Since these require manual review, it would be good to include an example for each one, so that there is less chance of differing interpretation

spopelka-dsac (changes requested):
Can you please use the existing `/db` directory, instead of the `subproject` directory, so it's clearer that this set of code pertains to the database? Please also update either the db readme or this readme with instructions for updating SQL code in the db and repo, so that folks know what they should do to be consistent

spopelka-dsac (approved):
Looks good to me! Thanks for making these changes, Fred!

PR #15, opened 2025-07-16
Creating a new provider table

spopelka-dsac:
## What has changed

- Creates a new provider table that references the individual table, instead of having the NPI in the individual table. This makes it more straightforward for the API to determine which individuals to include in the Practitioner endpoint and would allow for accessing specific Practitioners by NPI instead of an arbitrary other integer. 
- Adds in a gender field, which is required by FHIR.
- Removes some files that are not production ready from main

## How to test
- [ ] Does this change seem reasonable?


db/sql/create_table/merged_ndh_create_table.sql
```diff
@@ -227,10 +227,15 @@ CREATE TABLE ndh.individual (
     id SERIAL PRIMARY KEY,
     ssn VARCHAR(10)   DEFAULT NULL,
     sex_code CHAR(1)  DEFAULT NULL,
```
ftrotter-gov, line 229: you are right, gender is the one that is required.. so lets remove sex.

ftrotter-gov (changes requested):

db/sql/create_table/merged_ndh_create_table.sql
spopelka-dsac: Let's just keep both for now, it can just be empty

db/sql/create_table/merged_ndh_create_table.sql
```diff
@@ -227,10 +227,15 @@ CREATE TABLE ndh.individual (
...
-    npi bigint
+    gender_code CHAR(1),
+    birth_date DATE
 );
 
+CREATE TABLE ndh.provider (
+    npi BIGINT PRIMARY KEY, --references npi.npi
+    individual_id int --references individual.id
+)
+
```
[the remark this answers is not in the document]
spopelka-dsac: Thanks!

PR #16, opened 2025-07-18
Adding Foreign Keys for the Provider Tables

spopelka-dsac:
This PR makes the following changes:
* simplifies file naming in the `sql/` directory
* adds the foreign keys for the individual/provider tables
* changes the data type of the foreign key in the individual_to_nucc_taxonomy_code table, so that the data type matches the column that the fk references
* Removes the arbitrary id column in the nucc_taxonomy_code_to_medicare_provider_type


db/sql/schemas/individual_tables.sql
```diff
@@ -330,8 +330,8 @@ CREATE TABLE ndh.medicare_provider_type (
 
 --FRED: Is this part of the hydration or something we'll be storing like this long term?
 CREATE TABLE ndh.nucc_taxonomy_code_to_medicare_provider_type (
```
spopelka-dsac, line 332: @ftrotter-gov Can we just add the medicare_provider_type_id column to the nucc_taxonomy_code table and get rid of this additional table?

ftrotter-gov (approved):
I am trying to approve here...

PR #17, opened 2025-07-18
Sjp/practitioner endpoint

spopelka-dsac:
WE HAVE A WORKING PRACTITIONER ENDPOINT!

This PR does a ton of stuff:
- Renames the Django project to ndhfhir (to differentiate that this is the fhir api)
- Imports the db tables as python models
- Defines an adapter between the python models and the FHIR model for Practitioner
- Defines a ViewSet to list all Practitioners `http://localhost:8000/fhir/Practitioner` and display a single Practitioner `http://localhost:8000/fhir/Practitioner/<npi>`

### To Review
- Run the Django app and ensure that you can successfully hit both the Practitioner list endpoint and the single Practitioner endpoint (using npi 9999999999)
- Ensure that the adaptations generally make sense

### Single Practitioner
image

### Practitioner List
image


ftrotter-gov (approved):
Excellent Progress

PR #18, opened 2025-07-18
Adding the sql for the insert_practitioner function

spopelka-dsac:
## What Changed
This PR creates a sql function (`db/sql/functions/insert_practitioner.sql`) and associated documentation, for the purpose of doing a quick and dirty practitioner insert. If we decide it's useful to improve, even after we have data in the db, we can improve it, but until then it gets the job done. 

## How to test it
Follow the instructions in the readme to insert a dummy practitioner record

### DIFF
--- a/db/sql/functions/README.md
+++ b/db/sql/functions/README.md
@@ -1,0 +1,25 @@
+# Helper SQL Functions
+
+## insert_practitioner
+### Description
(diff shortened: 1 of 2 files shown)


ftrotter-gov (approved):
looks good.

spopelka-dsac:
> Just FYI there is an existing faker sub-project that does a pretty good job faking NPPES data if you ever need it: https://pypi.org/project/faker-healthcare-system/

Great resource!!! I'm going to whip up a quick helpful resourcres wiki and put this there so it doesn't get lost

PR #22, opened 2025-07-23
Adding Query Params and Pagination

spopelka-dsac:
This PR adds in query parameters for name, gender, and practitioner-type (nuuc display name), as well as pagination to handle large responses


decause-gov (approved):
I am mostly just here to prevent removing branch protection force pushes, but, LGTM +1 🚢

ftrotter-gov:
Got here as quick as I could!

PR #27, opened 2025-07-28
Doc Fixes

sachin-panayil:
This PR is meant to fix some document stuff within the repo.

- Removed repo-scaffolder templates
- Added a code.json
- Fixed contributor failing action bug


spopelka-dsac (approved):
Thanks for this @sachin-panayil!! LGTM. We need to update some of the other docs now that we have an official project name and more of a plan, so we'll probably do another pass at the code.json at that point, to ensure consistency.

PR #43, opened 2025-08-18
clarifying the table naming/schema

spopelka-dsac:
I didn't catch this at the time, but this change is a better way of specifying the schema. The other way (using `ndh.fips_county`) creates a table named `ndh.fips_county` in the public schema.


wbprice (approved):
I'm having trouble testing this locally but it _looks_ correct.

ftrotter-gov (approved):
given coming name change, a must have.

PR #47, opened 2025-08-20
Add Template for Pull Requests

IsaacMilarky:
## Problem

We currently do not have a template for PR descriptions.

## Solution

Add standard template from the OSPO.

## Result

We will now have a default template when creating a Pull Request into NPD.


spopelka-dsac (changes requested):
Can you please add a section for Jira ticket number, if applicable? Otherwise looks great!

spopelka-dsac (approved):
Thank you!

PR #48, opened 2025-08-20
automate running fixture data sql script

wbprice:
Add db migration scripts to docker compose, correct django environment variable name


spopelka-dsac:
Awesome, thanks!

PR #51, opened 2025-08-26
Adding terraform files

spopelka-dsac:
## Problem

The terraform code we used to set up the initial components was in a separate repo. (Shoutout to Pat Sier for having done the heavy lifting on the template)

## Solution

Moved the code to this repo

## Result

Now we have the files in this directory!

## Test Plan

N/A


wbprice (approved):
LGTM

PR #59, opened 2025-08-29
DRAFT: Build Remaining Tables

spopelka-dsac:
## Problem

The NDH schema is out of date and does not include the additional tables that are necessary to account for organizations.

## Solution

This creates a new NPD schema that updates the tables that exist in NDH and adds in the missing tables. This schema roughly follows the [Quick Database Diagrams ERD](https://app.quickdatabasediagrams.com/#/d/diFBJ4) with some divergence to account for bugs in that schema and the fact that that schema is simplified.

## Result

* Adds a PNG of the QDBD rough ERD
* Adds an NPD sql schema
* Removes confusing extra tables

## Test Plan

Look this over to let me know if anything seems glaringly problematic! Note: I will probably tweak it a bit here and there as I re-configure the FHIR api and data inserts to hook into the new schema.


spopelka-dsac:
@wbprice FYI I'll open a real PR after I update the FHIR API code accordingly, but would like a glance at the sql schema before I do so please!

PR #61, opened 2025-09-02
Configure S3 backend for Terraform

wbprice:
### Jira Ticket #
n/a

## Problem

Terraform, under specific circumstances, will persist secrets to `terraform.tfstate`, which is currently tracked by git. To avoid leaking secrets this way, we should track the backend using S3 instead of a local file.

## Solution

Configures an S3 backend in main.tf

## Result

Terraform tracks state using S3, rather than a developers local machine.

## Test Plan

`terraform plan` should return a minimal (or no) diffset.


infrastructure/main.tf
```diff
@@ -7,6 +7,13 @@ terraform {
       version = "~> 5.0"
     }
   }
+
+  backend "s3" {
+    bucket = "npd-terraform"
+    key = "terraform.tfstate"
+    region = "us-gov-west-1"
+    use_lockfile = true
```
spopelka-dsac, line 15: I got this error when running `terraform init`

image

spopelka-dsac (commented):
I got an error about the use_lockfile argument (ss commented on the offending line in main.tf), but otherwise lgtm

infrastructure/main.tf
wbprice: What version of terraform are you using?

PR #67, opened 2025-09-03
Fix the Action that Updates COMMUNITY.md with the Contributer Count

IsaacMilarky:
## Problem

The `contributers.yml` GitHub action workflow is failing

## Solution

Remove the upstream action that we were using to bypass branch protection rules and instead pass a PAT from @decause-gov into the checkout action. Also add @decause-gov as an exception to branch protection rules. 

## Test Plan

Here is the test [run](https://github.com/DSACMS/npd/actions/runs/17446051405)


spopelka-dsac (approved):
woot woot! Thank you!!

PR #73, opened 2025-09-09
Wbprice/npd 123 configure flyway migrations

wbprice:
## Problem

Using a database-first approach, we need a way to manage migrations that is not coupled to django. Flyway is an open source tool for managing database migrations that I've used on other CMS projects.

This PR _does not_ handle running migrations against the AWS database, but we could do something similar with Github Actions.

## Solution

- Add a flyway container spec to docker compose
- Configure flyway to run an initial database schema migration on docker compose up

## Result

Easier to contributors to start the API and load starting schema into the database

## Test Plan

docker compose up should start the database, THEN run migrations, THEN start the python app.

### DIFF
--- a/src/docker-compose.yml
+++ b/src/docker-compose.yml
@@ -17,7 +17,7 @@ services:
     image: 'flyway/flyway:10'
     command: migrate
     volumes:
-      - '../db/sql/migrations:/flyway/sql'
+      - '../db/sql:/flyway/sql'
     environment:
       FLYWAY_URL: jdbc:postgresql://${NPD_DB_HOST}:${NPD_DB_PORT}/${NPD_DB_NAME}
       FLYWAY_USER: ${NPD_DB_USER}


src/npdfhir/tests.py
```diff
@@ -16,7 +16,7 @@ def setup_databases(self, **kwargs):
 
         # Apply unmanaged tables
         with connection.cursor() as cursor:
-            cursor.execute(open("../db/sql/schemas/npd.sql").read())
+            cursor.execute(open("../db/sql/schemas/V1__initial_npd_schema.sql").read())
```
wbprice, line 19: I don't know yet how the python e2e tests interact with Postgres running in a docker container, talking through this now with @IsaacMilarky

src/README.md
```diff
@@ -16,7 +16,7 @@ Django backend that provides a FHIR API for accessing data from the npd database
 ### Local dev
 1. Ensure that either colima (if using macOS) or the docker service is running
 2. Create a `.env` file in this directory, following the template of the `.env_template` file
-    * n.b. ensure that NPD_DB_HOST is set to `host.docker.internal` if using a local postgres instance.
+    * n.b. ensure that NPD_DB_HOST is set to `db` if using a local postgres instance.
```
wbprice, line 19: There's probably some world where we want to be able to resolve the development database in AWS with this environment variable, so I left it in the directions.

PR #84, opened 2025-09-11
Add Fhir Validation Module

IsaacMilarky:
## Problem

We need to validate the schema of fhir data as well as the nucc, npi, and c80 codes. 

## Solution

Design a Python module to be used in the tests of the fhir data that we are going through. 

Use Pydantic, fhir.resources, and the database to validate the schema of these

## Results

Summary:

- Use pydantic to validate the schemas of fhir objects
- Create validator classes that can iterate through many python dictionaries and validate
- Create two subclasses that also have methods to check codes and npi
- Add a new table to hold the definitions of the c80 codes
- Add methods to validate the npi for practitioners
    -  New method to check the range of the given npi
    - New method to check the npi against the luhn algorithm 
- Add methods to test the given codes against our c80 and nucc tables

## Test Plan

Run `make test`

Results:
```

Running tests...
----------------------------------------------------------------------
./usr/local/lib/python3.13/site-packages/rest_framework/pagination.py:207: UnorderedObjectListWarning: Pagination may yield inconsistent results with an unordered object_list: <class 'npdfhir.models.EndpointInstance'> QuerySet.
  paginator = self.django_paginator_class(queryset, page_size)
........Not Found: /fhir/Endpoint/12300000-0000-0000-0000-000000000123/
......./usr/local/lib/python3.13/site-packages/rest_framework/pagination.py:207: UnorderedObjectListWarning: Pagination may yield inconsistent results with an unordered object_list: <class 'npdfhir.models.ClinicalOrganization'> QuerySet.
  paginator = self.django_paginator_class(queryset, page_size)
.....Not Found: /fhir/Organization/999999/
./usr/local/lib/python3.13/site-packages/rest_framework/pagination.py:207: UnorderedObjectListWarning: Pagination may yield inconsistent results with an unordered object_list: <class 'npdfhir.models.Provider'> QuerySet.
  paginator = self.django_paginator_class(queryset, page_size)
......Not Found: /fhir/Practitioner/999999/
.
----------------------------------------------------------------------
Ran 29 tests in 0.495s

OK

Generating XML reports...
```


src/npdfhir/fhir_checks.py
```diff
@@ -0,0 +1,151 @@
...
+    """
+
+    with connection.cursor() as cursor:
+        cursor.execute(query)
+        results = cursor.fetchall()
+
+    return parse_nucc_codes_into_dicts(results)
+
+
+def is_valid_npi_format(npi_value):
```
spopelka-dsac, line 32: This resolves to true for "0000000001" which is not a valid NPI. A safer check here would be to check that the NPI is an integer between 999999999 and 10000000000.

src/npdfhir/fhir_checks.py
```diff
@@ -0,0 +1,151 @@
...
+                self.verification_steps(entry)
+            except ValidationError as e:
+                errors.append(e)
+        
+        if len(errors) > 0:
+            print(f"Found {len(errors)} validation errors!")
+            #Requires python 3.11+
+            raise ExceptionGroup("Validation Errors Summary", errors)
+
+class FHIRValueSetVerifier(BaseVerifier):
```
spopelka-dsac, line 111: This is a good reminder to myself that we can be using our FHIR [codings](https://hl7.org/fhir/datatypes.html#Coding) more intelligently, to provide for automated verification across code sets. If we consistently use the FHIR ValueSets as systems, then we can automate ValueSet verification across the board. 


In practice, this looks like using the FHIR version of the taxonomy set instead of the NUCC version (http://hl7.org/fhir/us/ndh/STU1/ValueSet-IndividualAndGroupSpecialtiesVS.html), which would allow us to replace the '.html' at the end with '.json' to request a json representation of the ValueSet. With that, we'd be able to use a generalized function to check any code value against any code system by structuring the lookup to use the specification of the FHIR ValueSet json returns.

spopelka-dsac: So basically, instead of writing ValueSetVerifiers that are specific to specific CodeSets, we should have one that validates any ValueSet communicated using the FHIR ValueSet spec, and I'll create another ticket to update the NUCC system accordingly

IsaacMilarky: I gotcha, I think I might have some questions about when we fetch the taxonomy from hl7 and put it/ check it against the database but I think I understand the general principle.

backend/npdfhir/fhir_checks.py
```diff
@@ -0,0 +1,283 @@
...
+    Returns:
+        dict: Returns a mapping of the codes to what they correspond to in the db
+    """
+    codes = {}
+    for code_record in db_result:
+        codes[code_record[0]] = code_record[1]
+
+    return codes
+
+def get_nucc_codes_from_db():
```
spopelka-dsac, line 31: FWIW, we're already doing this using the ORM in cache.py, which is more robust and less redundant

backend/npdfhir/fhir_checks.py
```diff
@@ -0,0 +1,283 @@
...
+        return d * 2 if d < 5 else d * 2 - 9
+
+    digits_only = re.sub(r'\D', '', str(npi_value))
+
+    digits = [int(ch) for ch in digits_only[:10]]
+
+    total = sum(
+        transform(d) if i % 2 == 0 else d
+        for i, d in enumerate(digits)
+    ) + 24
```
spopelka-dsac, line 103: Why are you adding 24 here?

spopelka-dsac: Just answered my own question: https://www.cms.gov/Regulations-and-Guidance/Administrative-Simplification/NationalProvIdentStand/Downloads/NPIcheckdigit.pdf

Might be worth throwing that link in here for context

spopelka-dsac (changes requested):
What's your plan for incorporating this validation into django tests?

backend/npdfhir/fhir_checks.py
spopelka-dsac, line 1: small nit, but can we call this validators.py, so that it's clearer to someone coming in with no context what this is?

spopelka-dsac:
Can you please double check that the validators import without error after fixing this issue?

PR #89, opened 2025-09-15
Flyway retries connection if it fails

wbprice:
## Problem

Discovered a race condition where flyway would not wait until database was ready for connections before starting migrations on my local machine

## Solution

Added `-connectRetries=10` to migrate command.

## Result

Race condition gets a bandaid and migrations work as intended

## Test Plan

docker compose up, try to repro this error

image


spopelka-dsac (approved):
I was not receiving that error, so cannot repro, but this fix doesn't introduce any errors for me!

PR #90, opened 2025-09-16
Auto Reload for docker-compose

sachin-panayil:
## Problem

When running `/backend `with `docker-compose up --build`, you have to manually stop the server anytime you make code changes creating dev burden.

## Solution

Add volume mounts to bind your source code to the container.

## Result

Any changes to your code will be reflected in the server automatically meaning no more stopping and running loops.

## Test Plan

1. run `docker-compose up --build`
2. go to any endpoint
3. make changes
4. check to see if endpoint changes are reflected automatically

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 *.log
 *.pot
 *.pyc
+*.djcache
 __pycache__/
 local_settings.py
 db.sqlite3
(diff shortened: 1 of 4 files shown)


wbprice:
This works, but it drops a bunch of cache files in the PyCharm sidebar:
image
There's probably a way to configure the volumes such that `djcache` files don't end up in the `backend` directory?

spopelka-dsac (changes requested):
Minor changes. Otherwise looks great!

spopelka-dsac (approved):
Thanks!

PR #97, opened 2025-09-19
configure logging such that 500s and such are logged to stdout

wbprice:
http://ndh-alb-1866093491.us-gov-west-1.elb.amazonaws.com/fhir/Organization is responding with a 500 and I don't have enough logging to troubleshoot. I'm reading that when DEBUG is off, Django wants to email an admin the error.

image

This PR updates the logging configuration to include the stack trace in logs:

image

Eventually we want to instrument the API with Datadog or similar, but this is enough to start.

### DIFF
--- a/backend/app/settings.py
+++ b/backend/app/settings.py
@@ -14,6 +14,7 @@ from pathlib import Path
 from decouple import config
 import os
 import sys
+import logging
 
 # Detect if tests are being run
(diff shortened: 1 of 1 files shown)


wbprice:
More info on django logging config https://docs.djangoproject.com/en/5.2/howto/logging/

spopelka-dsac (approved):
lgtm

PR #100, opened 2025-09-22
NDH-219: Add addresses to the API

spopelka-dsac:
## Problem

We didn't have any address test data, so we were not exposing addresses through the API. 

## Solution

This PR loads processed address data from the ETL database (just from a CSV for now, as the address tables are going to change as Fred works on them), updates the Address serializer, Organization serializer, and Practitioner serializer to work with actual data, and updates/adds the FHIR-specified address query parameters.

TODO: Write tests

## Test Plan

The tests are currently borked until we get them working with docker and flyway. In the meantime, do the following: 
1. navigate to `/backend`
2. run `docker-ccompose up --build`
3. navigate to `http://localhost:8000/fhir/docs`
4. confirm that you see the added address query string parameters for `/Organization` and `/Practitioner`
5. filter by a few addresses/address parts to ensure that the results look as expected.

image

### DIFF
--- a/backend/npdfhir/serializers.py
+++ b/backend/npdfhir/serializers.py
@@ -246,13 +246,6 @@ class EndpointIdentifierSerialzier(serializers.Serializer):
     def to_representation(self, instance):
         endpoint_identifier = Identifier(
             use="official",
-            type=CodeableConcept(
-                coding=[Coding(
-                    system="http://terminology.hl7.org/CodeSystem/v2-0203",
-                    code="",  # value omitted for now
-                    display=""  # value omitted for now
-                )]
-            ),
             system=instance.system,
             value=instance.other_id,
             # TODO: Replace with Organization reference


backend/npdfhir/test.ipynb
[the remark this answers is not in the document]
spopelka-dsac: No, just bad repo hygiene, sorry!!

backend/npdfhir/views.py
```diff
@@ -87,15 +93,43 @@ def list(self, request):
...
+            if param == 'address-postalcode':
+                providers = providers.annotate(
+                    search=SearchVector(
+                        'individual__individualtoaddress__address__address_us__zipcode')
+                ).filter(search=value)
+            if param == 'address-use':
+                if value in addressUseMapping.keys():
+                    addressUse = addressUseMapping.toNPD(value)
+                    providers = providers.filter(
+                        individual__individualtoaddress__address_use_id=addressUse)
```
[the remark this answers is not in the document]
spopelka-dsac: good callout

backend/npdfhir/mappings.py
```diff
@@ -25,3 +31,13 @@ def toNPD(self, fhirValue):
         "O": "Other"
     }
 )
+
+addressUseMapping = Mapping(
+    {
+        1: "home",
+        2: "work",
+        3: "temp",
+        4: "old",
```
wbprice, line 40: Do `temp`, `old` have specific meanings in the source data that we want to expose to the end user?

backend/npdfhir/mappings.py
```diff
@@ -17,6 +17,12 @@ def toNPD(self, fhirValue):
         else:
             return self.mapping.inverse[fhirValue]
 
+    def keys(self, which='fhir'):
+        if which == 'npd':
+            return list(self.mapping.keys())
+        else:
+            return list(self.mapping.inverse.keys())
```
wbprice, line 24: I'm expecting to see `keys` get called with the `which` argument somewhere in this PR or in existing code, but didn't see it in GitHub search. Tell me more about this?

backend/npdfhir/models.py
```diff
@@ -26,8 +26,7 @@ class Meta:
 
 
 class AddressInternational(models.Model):
-    id = models.UUIDField(primary_key=True)
-    input_id = models.CharField(max_length=36, blank=True, null=True)
+    id = models.CharField(primary_key=True, max_length=10)
```
wbprice, line 29: What was `input_id` doing before and why is it OK to remove?

spopelka-dsac:
> Listing endpoints gives a 500: image

Endpoints are broken in main. Sachin opened https://github.com/DSACMS/npd/pull/102 to fix it

backend/npdfhir/mappings.py
spopelka-dsac: They're from a [FHIR ValueSet](https://hl7.org/fhir/valueset-address-use.html). I kind of doubt that we'll be using temp and old, but I loaded the full valueset just in case

spopelka-dsac:
Moving this convo here:
> @wbprice: Is filtering implemented yet? I tried a search for providers in VA and got mostly VA providers but one in Florida


> @spopelka-dsac: That [org](https://npiregistry.cms.hhs.gov/provider-view/1124021720) has multiple addresses, one of which is in VA. The search is looking across all the addresses. I was working off of the base FHIR spec, which is much more restrictive on displaying addresses (single contact address). However, I'm noticing now that the NDH FHIR spec allows for displaying a list of addresses, so I'll update to that! (edited)

It's going to be a non-trivial lift to update the Organization validator to not error out at a list of addresses for Organizations. I opened this ticket: https://jiraent.cms.gov/browse/NDH-238 to track that work

PR #102, opened 2025-09-23
Fixing broken endpoint due to validation 

sachin-panayil:
## Problem

Endpoint endpoint was failing because I left the Identifier resource empty which failed validation tests

## Solution

Implemented a default of ANON for Code and Display within EndpointIdenfifierSerializer

## Result

Endpoint endpoint works as it should

* I just randomly chose ANON but can change it to what makes more sense as a default code
* I will implement this into testing for this endpoint but i wonder if we should have CI tests that run these validation checks every time someone is working on the API

## Test Plan

Run docker-compose up --build and hit endpoint.

### DIFF
--- a/backend/npdfhir/serializers.py
+++ b/backend/npdfhir/serializers.py
@@ -253,13 +253,6 @@ class EndpointIdentifierSerialzier(serializers.Serializer):
     def to_representation(self, instance):
         endpoint_identifier = Identifier(
             use="official",
-            type=CodeableConcept(
-                coding=[Coding(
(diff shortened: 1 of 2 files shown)


backend/npdfhir/serializers.py
```diff
@@ -256,8 +256,8 @@ def to_representation(self, instance):
             type=CodeableConcept(
```
spopelka-dsac, line 256: Can you just remove type please?

spopelka-dsac (changes requested):

spopelka-dsac (approved):

PR #103, opened 2025-09-24
Working tests on container

IsaacMilarky:
## Problem

Currently, tests only work on a bare-metal postgres instance.

## Solution

Make sure that the tests are run as docker native. They should be able to be ran on the django-web container and have the results saved as xml on the local machine. 

## Result

Summary

* Configure Django tests to use preexisting database instead of creating a new one.  
* Replace custom DiscoverRunner logic with a test database that is prepopulated with test data + schema using flyway 
* Add `unittest-xml-reporting` as a dependency for xml reporting

## Test Plan

In the backend directory: 

First, run `make dev-up`
Then, run `make test`


backend/docker-compose-test.yml
```diff
@@ -0,0 +1,63 @@
...
+      - "${NPD_DB_PORT}:5432"
+    networks:
+      - npd
+    tmpfs:
+      - /var/lib/postgresql/data:rw,size=200m
+    healthcheck:
+      test: ["CMD-SHELL", "pg_isready -U ${NPD_DB_USER} -d test_${NPD_DB_NAME}"]
+      interval: 5s
+      timeout: 5s
+      retries: 10
```
abachman-dsac, line 19: Using a duplicate implementation of the database service exclusively for testing feels like an anti-pattern on a project like this. I would like to see us relying on the default compose.yml DB service with the test suite setup code providing the unique `test_${NPD_DB_NAME}`.

A common pattern I've enjoyed is using a prefix or suffix--as demonstrated here--for databases in each environment the application executes in to help make sense of setup, logging, error traces, etc.

backend/docker-compose-test.yml
```diff
@@ -0,0 +1,63 @@
...
+    environment:
+      FLYWAY_URL: jdbc:postgresql://db-test:5432/test_${NPD_DB_NAME}
+      FLYWAY_USER: ${NPD_DB_USER}
+      FLYWAY_PASSWORD: ${NPD_DB_PASSWORD}
+    depends_on:
+      db-test:
+        condition: service_healthy
+    networks:
+      - npd
+  django-test:
```
abachman-dsac, line 34: I think this is an okay place to start, but I'm curious to see how maintainable it will be over the long run. 

Tests kind of want to run in an environment almost-but-not-exactly like the environment the code expects to run in, which means we can get away with a lot of reuse in our docker compose service definitions.

backend/docker-compose-test.yml
```diff
@@ -0,0 +1,63 @@
...
+      NPD_DB_USER: ${NPD_DB_USER}
+      NPD_DB_PASSWORD: ${NPD_DB_PASSWORD}
+      NPD_DB_HOST: db-test
+      NPD_DB_PORT: "5432"
+    env_file:
+      - .env
+    command: >
+      sh -c "python manage.py test"
+    volumes:
+      - ./artifacts:/tmp
```
abachman-dsac, line 61: nit (non-blocking): writing a compose.yml file with no line breaks between blocks makes the code read a little tight. We don't have a style linter for our yaml, so this is a fine style, but it feels a little claustrophobic.

IsaacMilarky:
> With the changes to `backend/Makefile` to remove `sudo`, I think the current branch is just about good to merge. Particularly if it puts the project in a position to have the test suite run in CI / GitHub Actions.
> 
> The last nit for me is kind of a "developer affordances" thing. I notice that I'm having a bit of difficulty running `docker compose up` after running `make test` because the two db servers use the same port. Similarly, it's hard to go the other way and run `docker compose up` to get a dev env working after successfully running `make test` since the new test command leaves a test-only postgres service--`db-test`--running on the same port that the `docker-compose.yml` `db` service wants to use.
> 
> The only thing I would really like to see to give an enthusiastic +1 is a touch more developer documentation in the PR description demonstrating an ideal workflow for moving back and forth--as a developer at the command-line--from testing to development in the `backend/` sub-directory.
> 
> _PS_: There also seems to be some docker compose errors that come out of switching between compose files which use different networks having the same name and which also share a service (`db-migrations`) that does two different things. But I'm not sure if that's true for other users or just a fluke of my system setup.

Sounds good! I have changed the branch to only use one `docker-compose.yml` file. Now, it works by first running the development containers with `make dev-up` and then running `make test`. To shutdown the dev containers then run `make dev-down`

PR #105, opened 2025-09-24
NDH-178: Fix Incorrect Content-Type

sachin-panayil:
## Problem

The Content-Type header for our FHIR APIs should return “application/fhir+json” and we have a test that is currently failing because it does not. The header should be fixed so the test passes.

Our original solution of defining the content type within the response object, `response["Content-Type"] = "application/fhir+json"`,  was getting overridden by default renderer when a HTTP request was being sent.

## Solution

Added custom renderers that inherits from `JSONRenderer` so that we can manually set our Content-Type.

## Result

The Content-Type for all requests are now `application/fhir+json`. This also is compatible with the Browsable API since we added `BrowsableAPIRenderer`

### Notes 
- I didnt get to run the test since im waiting on Isaacs changes however im pretty sure they will pass 
- Screenshot of Swagger docs showing headers and proper Content-Type
Screenshot 2025-09-26 at 3 48 33 PM

## Test Plan

- Hit any endpoint after running `docker-compose up --build` and check the headers
- View the frontend by going to `http://localhost:8000/fhir` on any browser after running docker

### DIFF
--- a/backend/npdfhir/views.py
+++ b/backend/npdfhir/views.py
@@ -235,7 +235,6 @@ class FHIRPractitionerViewSet(viewsets.ViewSet):
 
         # Set appropriate content type for FHIR responses
         response = paginator.get_paginated_response(bundle.data)
-        response["Content-Type"] = "application/fhir+json"
 
         return response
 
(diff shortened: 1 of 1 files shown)


abachman-dsac (approved):
Thank you for updating the PR description! 

I was also able to use the swagger docs browser at http://localhost:8000/fhir/docs and find an API call that demonstrated the new headers on a default build of the development service: 

```console
$ curl -v -X 'GET' \
  'http://localhost:8000/fhir/Endpoint/?page_size=10' \
  -H 'accept: application/fhir+json' 2>&1 | \grep Content-Type:
< Content-Type: application/fhir+json
```

👍

PR #107, opened 2025-09-30
NDH-223: Adding Endpoint Endpoint Tests

sachin-panayil:
## Problem

The newly added Endpoint endpoint did not have tests associated with them to ensure they are valid when ran.

## Solution

Write a basic suite of tests based on the tests that were created for previous endpoints.

## Result

Endpoint endpoint now has a test suite that checks for basic functionality and happy paths which run successfully as show in screenshot below

Screenshot 2025-10-01 at 1 03 24 PM

#### Some important notes regarding the summary line:
* We should consolidate some of these tests somehow as a lot of the test suites are testing for the same thing
* I keep thinking about fixtures vs factories and what would make the most sense for an API like this

## Test Plan
- run `docker-compose up --build`
- run `docker-compose exec django-web python manage.py test`

### DIFF
--- a/backend/npdfhir/tests.py
+++ b/backend/npdfhir/tests.py
@@ -3,8 +3,9 @@ from rest_framework import status
 from rest_framework.test import APITestCase, APIClient
 from django.test.runner import DiscoverRunner
 from django.db import connection
-from .cache import cacheData
-# I can't explain why, but we need to import cacheData here. I think we can remove this once we move to the docker db setup
+from .cache import cacheData # I can't explain why, but we need to import cacheData here. I think we can remove this once we move to the docker db setup
+from fhir.resources.bundle import Bundle
+from pydantic import ValidationError
 
 def get_female_npis(npi_list):
     """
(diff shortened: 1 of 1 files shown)


backend/npdfhir/tests.py
```diff
@@ -24,6 +24,123 @@ def get_female_npis(npi_list):
...
+        self.list_url = reverse("fhir-endpoint-list")
+
+    def test_list_default(self):
+        response = self.client.get(self.list_url)
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(response["Content-Type"], "application/fhir+json")
+        self.assertIn("results", response.data)
+    
+    def test_list_returns_fhir_bundle(self):
```
IsaacMilarky, line 39: I would validate that the result is correct by using the Pydantic objects provided by fhir.resources. 

Something like:
```python
from fhir.resources.bundle import Bundle

try:
    _ = Bundle(data_dict_to_validate)
except ValidationError as e:
    #handle  me 
```

IsaacMilarky (changes requested):
Overall looks great! One small nitpick is that I would use the fhir.resources package with the built in Pydantic validation for consistency.

backend/npdfhir/tests.py
sachin-panayil: good call out, we should update these in all the tests later

PR #109, opened 2025-10-07
NDH-249: Update the Organization and Practitioner to use organization_id 

sachin-panayil:
## Problem
The FHIR API was using NPI as the primary identifier for both Organization and Practitioner resources, limiting queries to only organizations with NPIs. This prevented access to organizations without NPIs and created inconsistency in the API design.

## Solution
- Changed Organization to use `organization.id` instead of NPI as primary identifier
- Changed Practitioner to use `individual.id` instead of NPI as primary identifier
- Implemented identifier search functionality:
  - General search (`?identifier=value`) - searches across all identifier types (NPI, EIN, Other IDs)
  - Specific search (`?identifier=system|value`) - searches within a specific system
  - Created `parse_identifier()` helper function to handle format

## Result
Endpoints now support the followng
- `/Organization/{uuid}` and `/Practitioner/{uuid}`
- `?identifier=12345` - searches through all systems
- `?identifier=NPI|12345` - searches only NPIs
- `?identifier=EIN|{uuid}` - searches only EINs
- `?identifier=cmsMBI|12345` - searches only OtherIDs

## Test Plan
_Not sure if I can paste the test data here but you can find them if do /Organization/ and replace the respective fields with the ones you find within response. I am using Postman to test these but `curl` should be fine too_

**1. Retrieve search**
- `GET /Organization/{valid-uuid}` = pass with valid resource returned
- `GET /Practitioner/{valid-uuid}` = pass with valid resource returned
- `GET /Organization/999999` = fails with error bc invalid uuid

**2. General search**
- `GET /Organization/?identifier={npi}` = pass with valid Bundle returned with matching NPI
 - `GET /Organization/?identifier={ein}` = pass with valid Bundle returned with 0 Resources since we dont have test EIN loaded in database

**3. Specific searches**
- `GET /Organization/?identifier=NPI|{npi}` = pass with valid Bundle returned with matching NPI
- `GET /Organization/?identifier=EIN|{ein}` = pass with valid Bundle returned with 0 Resources since we dont have test EIN loaded in database
- `GET /Organization/?identifier=1|{otherID}` = pass with valid Bundle returned with matching otherID field
- `GET /Practitioner/?identifier=NPI|{npi}` = pass with valid Bundle returned with matching NPI

### DIFF
--- a/backend/npdfhir/views.py
+++ b/backend/npdfhir/views.py
@@ -12,7 +12,7 @@ from uuid import UUID
 from .models import Provider, EndpointInstance, ClinicalOrganization, Organization
 from .serializers import PractitionerSerializer, OrganizationSerializer, BundleSerializer, EndpointSerializer
 from .models import Provider, EndpointInstance, ClinicalOrganization
-from .serializers import PractitionerSerializer, ClinicalOrganizationSerializer, BundleSerializer, EndpointSerializer
+from .serializers import PractitionerSerializer, OrganizationSerializer, BundleSerializer, EndpointSerializer
 from .mappings import genderMapping, addressUseMapping
 from .renderers import FHIRRenderer
 from drf_yasg.utils import swagger_auto_schema


backend/npdfhir/views.py
```diff
@@ -40,6 +42,20 @@ def createFilterParam(field: str, display: str = None, enum: list = None):
     return param
 
 
+def parse_identifier(identifier_value):
+    """
+    Parse an identifier search parameter that should be in the format of "value" OR "system|value"
+    
+    Not sure the best way to document this but currently the only system are NPI, EIN or OtherID which could be anything...
```
spopelka-dsac: I'm comfortable with us just supporting `NPI | <npi>` for now. Do you mind simplifying it and then we can expand as needed, if needed?

backend/npdfhir/views.py
```diff
@@ -145,6 +160,7 @@ class FHIRPractitionerViewSet(viewsets.ViewSet):
     @swagger_auto_schema(
         manual_parameters=[
             page_size_param,
+            createFilterParam('identifier', 'format: value OR system|value -> 12345567 OR NPI|12345567'),
```
spopelka-dsac: nit, but since NPI is the only system we support atm, I think it's clearer to say something like "format: value (for any type of identifier) OR NPI|value (if searching for an NPI) -> 12345567 OR NPI|12345567"

backend/npdfhir/views.py
```diff
@@ -298,12 +388,36 @@ def list(self, request):
...
+                    else: # general identifier search requested
+                        try:
+                            queries = Q(clinicalorganization__npi__npi=int(identifier_id))
+                        except (ValueError, TypeError):
+                            pass
+
+                        if queries is None: # force return no results 
+                            queries = Q(pk__isnull=True)
+                    
+                    organizations = organizations.filter(queries).distinct()
```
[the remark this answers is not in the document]
sachin-panayil: fixed this by just initializing an emptyquery set in the beginning of block

backend/npdfhir/views.py
```diff
@@ -298,12 +388,36 @@ def list(self, request):
...
+                        if system == 'NPI':
+                            try:
+                                queries = Q(clinicalorganization__npi__npi=int(identifier_id))
+                            except (ValueError, TypeError):
+                                pass
+
+                        if queries is None: # force return no results 
+                            queries = Q(pk__isnull=True)
+
+                    else: # general identifier search requested
```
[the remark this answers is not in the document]
sachin-panayil: reverting back to old logic since using searchvector here might not work the best due to type errors

flyway/sql/inserts/R__test_non_clinical_organization.sql
abachman-dsac: Not a pressing issue for this PR, but a note for future test refactoring work: 

If this database setup is exclusively for testing, it should be part of the test suite in `setUp` instance methods as close to the relevant `TestCase` as possible. 

If it's data that won't be modified by any tests, it can be put in a `setUpClass` or `setUpTestData` data class method on the `TestCase`.

If it's data that is needed by multiple tests, we could put it in a test-suite level setup method.

Best solution is likely to be relying on [Django's built-in support for test fixtures](https://docs.djangoproject.com/en/5.2/topics/db/fixtures/) and then moving this data to a `npdfhir/fixtures/....json` file.

PR #113, opened 2025-10-15
[NDH-241] Landing page MVP

abachman-dsac:
## Landing Page: Styled MVP for the npd frontend application landing page

- [Jira Ticket NDH-241](https://jiraent.cms.gov/browse/NDH-241)

## Problem

Implement the Provider Search MVP landing page.

Figma screenshot:

image

## Solution

Introduce [CMS Design System](https://design.cms.gov/) and [USWDS](https://designsystem.digital.gov/) to implement the initial landing page.

## Result

Works On My Machine™

localhost_3000_


sachin-panayil (commented):
tested and works on my machine after running `docker compose up --build`. how do i test the FHIRPoc? i dont see a route for it but maybe im missing something. also, i wonder if we should make the header logo and title a button that navigates you back to landing page. otherwise looks great 🔥 

one things i did notice was when tabbing through `For Developers` submenu, it blocks out the text. this is not important for this months deadline but good to keep note of it.

Screenshot 2025-10-16 at 5 03 02 PM

PR #118, opened 2025-10-17
NDH-226: Add local and remote container hostnames to ALLOWED_HOSTS

tspecht-cms:
## Problem

https://medium.com/django-unleashed/djangos-allowed-hosts-in-aws-ecs-369959f2c2ab

Health checks will fail with a status 400 because the Host header of the request during health check is set to the containers private IP address. This IP address needs to be added to the ALLOWED_HOSTS list in Django’s setting.

## Solution

- uses custom middleware to bypass ALLOWED_HOSTS check for the `/fhir/healthCheck` endpoint only

## Result

Healthcheck requests from the load balancer will not longer result in 400 response and errors that the host is not allowed on the healthCheck endpoint

## Test Plan

Testing running the service locally using the middleware shows good results. This is with all ALLOWED_HOSTS lists empty, so that the localhost addresses are not explicitly allowed in Django.

Before:

image

After:

image

### DIFF
--- a/backend/npdfhir/middleware.py
+++ b/backend/npdfhir/middleware.py
@@ -1,0 +1,10 @@
+from django.http import HttpResponse
+
+class HealthCheckMiddleware:
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        if request.path == '/fhir/healthCheck':
+            return HttpResponse('healthy')
+        return self.get_response(request)


tspecht-cms:
@wbprice I appreciate your approach to exempt the health endpoint from enforcement of the allowed hosts list. I found a built-in Python tool (csrf_exempt) that does this and I wonder what you think of using this on the health endpoint instead of the middleware.

PR #120, opened 2025-10-20
NDH-203 Add PractitionerRole and Location Endpoints

spopelka-dsac:
## Problem

The current FHIR API has Organization endpoints and Practitioner endpoints, but no way to portray a relationship between the two. 

## Solution

This PR introduces a [PractitionerRole](https://build.fhir.org/ig/HL7/fhir-us-ndh/StructureDefinition-ndh-PractitionerRole.html) endpoint, which exposes Provider/Organization/Location relationship information. Since I was already modifying the location tables, I also threw in the Location endpoint, too. Note: the prior ERD contained composite foreign keys, which are not supported in Django, so this PR also removes those.

## Result

* Adds a new migration file to incorporate database changes to better support the PractitionerRole endpoint
* Updates the sample data insert file to include some re-structured sample data to better accommodate the new database schema
* Adds a link to a LucidChart with a new ERD (the AWS outage is preventing me from downloading a PNG)
* Updates backend/npd/models.py to reflect the new database schema
* Adds a Location serializer and a PractitionerRole serializer
* Adds a Location ViewSet and a PractitionerRole ViewSet
* Adds some basic tests for the Location ViewSet and PractitionerRole ViewSet

## Test Plan

1. navigate to `/backend`
2. `make test setup`
3. The tests should pass except for the not found errors, and pagination warnings, both of which are being fixed in other PRs
4. run `make setup` and `make up`
5. navigate to `localhost:8000/fhir/` and you should see that Location and PractitionerRole have been added to the list of endpoints
6. navigate to `localhost:8000/fhir/docs/` and you should see that there is documentation for the Location and PractitionerRole endpoints. You can also query them here.

Known next steps: 
* incorporate better endpoint sample data and add in the endpoint components
* Expand the query string parameters

### DIFF
--- a/backend/npdfhir/router.py
+++ b/backend/npdfhir/router.py
@@ -24,6 +24,13 @@ class OptionalSlashRouter(DefaultRouter):
 
 
 router = OptionalSlashRouter()
-router.register(r"Practitioner", views.FHIRPractitionerViewSet, basename="fhir-practitioner")
-router.register(r"Organization", views.FHIROrganizationViewSet, basename="fhir-organization")
-router.register(r"Endpoint", views.FHIREndpointViewSet, basename="fhir-endpoint")
+router.register(r"Practitioner", views.FHIRPractitionerViewSet,
+                basename="fhir-practitioner")
+router.register(r"Organization", views.FHIROrganizationViewSet,
+                basename="fhir-organization")
+router.register(r"Endpoint", views.FHIREndpointViewSet,
+                basename="fhir-endpoint")
+router.register(r"PractitionerRole", views.FHIRPractitionerRoleViewSet,
+                basename="fhir-practitionerrole")
+router.register(r"Location", views.FHIRLocationViewSet,
+                basename="fhir-location")


backend/npdfhir/serializers.py
```diff
@@ -374,6 +383,57 @@ def to_representation(self, instance):
...
+        representation = super().to_representation(instance)
+        practitioner_role = PractitionerRole()
+        practitioner_role.id = str(instance.id)
+        practitioner_role.active = instance.active
+        practitioner_role.practitioner = Reference(
+            reference=request.build_absolute_uri(f'Practitioner/{instance.provider_to_organization.individual_id}').replace('PractitionerRole/', ''))
+        practitioner_role.organization = Reference(
+            reference=request.build_absolute_uri(f'Organization/{instance.provider_to_organization.organization_id}').replace('PractitionerRole/', ''))
+        practitioner_role.location = [Reference(
+            reference=request.build_absolute_uri(f'Location/{instance.location.id}').replace('PractitionerRole/', ''))]
```
abachman-dsac: use of trailing slashes here would be helpful until we've sorted the trailing vs no-trailing issue in routing.

backend/npdfhir/serializers.py
```diff
@@ -374,6 +383,57 @@ def to_representation(self, instance):
...
+        model = ProviderToOrganization
+
+    def to_representation(self, instance):
+        request = self.context.get('request')
+        representation = super().to_representation(instance)
+        practitioner_role = PractitionerRole()
+        practitioner_role.id = str(instance.id)
+        practitioner_role.active = instance.active
+        practitioner_role.practitioner = Reference(
+            reference=request.build_absolute_uri(f'Practitioner/{instance.provider_to_organization.individual_id}').replace('PractitionerRole/', ''))
```
abachman-dsac: I may be missing something in my local setup, but it appears that the use of a UUID here breaks the logic in `FHIRPractitionerViewSet#retrieve` which wants to immediately wrap the given parameter with `int()` 

```python
    def retrieve(self, request, pk=None):
        """
        Return a single provider as a FHIR Practitioner resource
        """
        provider = get_object_or_404(Provider, pk=int(pk))         # <======
```

This is causing 500s for me whenever I visit one of the URLs generated by this endpoint.

For example, if I get a list of 1 PractitionerRole: 

```js
    ...,
    "entry": [{
        "fullUrl": "PractitionerRole/3ac7bd1e-a698-4905-9731-ca650de2dcb0",
        "resource": {
          "resourceType": "PractitionerRole",
          "id": "3ac7bd1e-a698-4905-9731-ca650de2dcb0",
          "active": true,
          "practitioner": {
            "reference": "http://localhost:8000/fhir/Practitioner/00c324a2-d978-44ca-907d-c054f8a59141"
          },
          "organization": {
            "reference": "http://localhost:8000/fhir/Organization/6cd0e991-7ee0-47fe-897e-a06a2cd5fdc3"
          },
          "location": [
            {
              "reference": "http://localhost:8000/fhir/Location/ad514df2-94b3-40dc-b129-10b1a62cfed1"
            }
          ]
        }
      }
```

Visiting any of the generated URLs (after appending a trailing slash ;)) causes a `ValueError` response:

Image

```
ValueError at /fhir/Location/ad514df2-94b3-40dc-b129-10b1a62cfed1/
invalid literal for int() with base 10: 'ad514df2-94b3-40dc-b129-10b1a62cfed1'
```

I think I'd need to know what's broken on my machine to review.

abachman-dsac (commented):
I have a question about the expected behavior of the single record URLs generated as part of the listing APIs.

backend/npdfhir/views.py
```diff
@@ -362,3 +455,114 @@ def retrieve(self, request, pk=None):
...
+        return response
+
+    def retrieve(self, request, pk=None):
+        """
+        Return a single provider as a FHIR Practitioner resource
+        """
+        location = get_object_or_404(Location, pk=pk)
+
+        location = LocationSerializer(
+            location, context={"request": request})
```
abachman-dsac: nit: i'm vaguely unnerved by the variable shadowing here. 

is this a common naming pattern with rest_framework viewsets?

how would you feel about `serialized_location = LocationSerializer(...)` ?

backend/npdfhir/urls.py
```diff
@@ -1,12 +1,12 @@
-from django.urls import path, include
+from debug_toolbar.toolbar import debug_toolbar_urls
+from django.urls import include, path
+from drf_yasg import openapi
+from drf_yasg.views import get_schema_view
+from rest_framework.renderers import JSONOpenAPIRenderer
 from rest_framework.routers import DefaultRouter
 from rest_framework.schemas import get_schema_view
```
abachman-dsac:
```suggestion
from debug_toolbar.toolbar import debug_toolbar_urls
from django.urls import include, path
from drf_yasg import openapi
from drf_yasg.views import get_schema_view
from rest_framework.routers import DefaultRouter
```

It looks like there was still an unused import and also an imported method with the same name as the one you intended.

abachman-dsac (changes requested):
one change needed in urls.py imports

PR #122, opened 2025-10-21
[NDH-304] Cleanly handle trailing slashes in Django paths

abachman-dsac:
## backend routing: Handle optional trailing slashes in our backend paths

[Jira Ticket NDH-304](https://jiraent.cms.gov/browse/NDH-304)

## Problem

We are annoyed by having to carefully type `/` onto the end of URLs we are using during development. 

We are also annoyed by having to carefully generate trailing slashes for all resource URLs in our API responses.

It would be nice if `localhost:8000/fhir` and `localhost:8000/fhir/` both went to the same place and it didn't break anything.

Our goal is that the following routing table would be accurate:

| path | view | 
| --- | --- | 
| `/fhir` | `"api-root"` |
| `/fhir/` | `"api-root"` |
| `/fhir/Endpoint` | `"fhir-endpoint-list"` |
| `/fhir/Endpoint/` | `"fhir-endpoint-list"` |
| `/fhir/Endpoint/12345` | `"fhir-endpoint-detail"` |
| `/fhir/Endpoint/12345/` | `"fhir-endpoint-detail"` |
| `/fhir/docs` | `"schema-swagger-ui"` |
| `/fhir/docs/` | `"schema-swagger-ui"` |
| `/` | `"provider_directory:index"` |
| `/whatever/npd/needs/` | `"provider_directory:index_with_path"` |

## Solution

Turns out to be a little tricky, but solvable.

There are a few situations conspiring to make this difficult for us. 

1. We use `include()` routing, like `path('fhir/, include('npdfhir.urls'))`, which is not amenable to regex paths with optional trailing slashes. 
2. We rely on a fall-through empty path matcher for the `rest_framework.DefaultRouter` "api-root" page.
3. We mounted other paths in addition to the `DefaultRouter` in the `npdfhir` app.
4. Django's `APPEND_SLASH`--which is enabled by default--does not work with nested url paths having optional trailing slashes.
5. Same with `rest_framework.DefaultRouter(trailing_slash=boolean)`

I added notes to `backend/app/urls.py` describing the challenges in a comment and added some new routing-specific tests to `backend/api/tests/test_routing.py` to ensure there is coverage on the setup.

The primary changes are: 

- extract our `rest_framework.DefaultRouter` implementation into its own module in `backend/npdfhir/router.py`
- implement a simple custom router wrapping `DefaultRouter` which accepts optional trailing slashes on all endpoints
- introduce a separate `path()` definition pointing requests to `/fhir` (no trailing slash) directly at the DefaultRouter api-root view

I also ran `ruff format` on the files I created or substantially updated.

## Result

Going forward, the backend application will be slightly more liberal in the paths it will accept while still routing correctly / in the way we expect it to. Existing routing for URLs already in the wild will be unaffected.

The application will continue to be particular in the routes it generates using Django's `reverse(...)` helper method.

- ✅ `/fhir/Endpoint` and `/fhir/Endpoint/` point to the same view  
- ✅ `reverse("fhir-endpoint-list")` always returns `"/fhir/Endpoint"`

## Test Plan

There are new tests in `backend/app/tests/`, they can be run alone with: 

```sh
python manage.py test app.tests
```

Or as part of the whole suite.

### DIFF
--- a/backend/app/settings.py
+++ b/backend/app/settings.py
@@ -84,6 +84,7 @@ CORS_ALLOW_ALL_ORIGINS = True
 CORS_ALLOWED_METHODS = ['GET']
 
 ROOT_URLCONF = 'app.urls'
+APPEND_SLASH = True # this is default, but we're making sure it's explicit
 
 TEMPLATES = [
     {
--- a/backend/app/tests/test_routing.py
+++ b/backend/app/tests/test_routing.py
@@ -1,0 +1,81 @@
+from django.urls import resolve, reverse
+from django.test import TestCase
+
+
+class ProviderDirectorySpaRouting(TestCase):
+    """
+    Ensure that provider_directory routing is agnostic to paths.
+    """
+
+    def test_homepage_url_resolves_to_correct_view(self):
+        resolver = resolve("/")
+        self.assertEqual(resolver.view_name, "provider_directory:index")
+        self.assertEqual(resolver.func.__name__, "index")
+
+    def test_detail_url_resolves_with_path(self):
+        resolver = resolve("/search/123/")
+        self.assertEqual(resolver.view_name, "provider_directory:index_with_path")
+        self.assertEqual(resolver.kwargs["path"], "search/123/")
+
+    def test_reverse_index_returns_path(self):
+        path = reverse("provider_directory:index")
+        self.assertEqual(path, "/")
+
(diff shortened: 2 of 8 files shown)


spopelka-dsac (approved):
Looks great to me and works well!

PR #127, opened 2025-10-22
NDH-326 Clarify FHIR Version for Base Validator

spopelka-dsac:
…it that we're using FHIR v4

## Problem

The meeting with Alberto this morning, in which we discussed fhir profiles and validators, made me. realize that the code was conflating a few implementation guides and FHIR versions. We need to be more explicit/consistent about which base FHIR version we're using and where we're layering on top of it.

## Solution

This PR is a starting point, that takes us down to brass tacks with a v4 FHIR base implementation. From here, we can layer in extensions on the validators that make it clearer where we're relying on other fhir profiles/ IGs. 

## Result

* Explicitly imports the FHIR v4 base validator from `fhir.resources` instead of the most recent version
* Makes some tweaks to bring the serializers into conformance with the v4 base FHIR spec
* Updates the tests accordingly

## Test Plan

1. navigate to `/backend`
2. run `make test`
3. all tests should pass!

Note: This is a little messy but will get cleaned up as we layer the explicit us core/ ndh/ plan net validators in

### DIFF
--- a/backend/npdfhir/serializers.py
+++ b/backend/npdfhir/serializers.py
@@ -1,21 +1,21 @@
 import sys
 
 from django.urls import reverse
-from fhir.resources.address import Address
-from fhir.resources.bundle import Bundle
-from fhir.resources.codeableconcept import CodeableConcept
-from fhir.resources.coding import Coding
-from fhir.resources.contactpoint import ContactPoint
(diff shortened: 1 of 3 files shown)


abachman-dsac (approved):
no objections 👍 

good to move towards clarity

PR #128, opened 2025-10-22
Wbprice/GitHub actions runner

wbprice:
## Problem

We need to configure an instance and associated roles for deploying infrastructure to AWS in a self-hosted runner to keep using GitHub Actions

## Solution

- Specified deploy-specific infrastructure 
- documented manual steps

## Result

dev environment is updated when deploys are made

## Test Plan

dev environment should deploy using a self-hosted runner (by manually starting deploy to dev task in the other repo)

### DIFF
--- a/infrastructure/envs/dev/main.tf
+++ b/infrastructure/envs/dev/main.tf
@@ -98,10 +98,11 @@ module "ecs" {
 module "fhir-api" {
   source = "../../modules/fhir-api"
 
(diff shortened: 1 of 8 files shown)


wbprice:
The dev deploy gets pretty far right now, but fails at the last step:
image

spopelka-dsac (approved):
lgtm

PR #129, opened 2025-10-23
ndh-286: add database check to healthCheck endpoint; update unit test

tspecht-cms:
## Problem

The health check endpoint does not check database connectivity, except when the application is initially launched

## Solution

Add database check to health check endpoint so it runs whenever the endpoint is called

## Result

Used built-in Django `connection.cursor()` method to run a simple SQL command `SELECT 1` against the database to check for a response. This is a standard method for checking connectivity to a database in a situation where a lightweight response is all that is needed. 

The code is added in the middleware, so that it maintains the exemption from the allowed hosts check that was set up for the health check endpoint in ticket #226.

## Test Plan

Updated the health check endpoint unit test to reflect the new response from the the endpoint, which looks like:

image

Tested locally with the database running, and with the database stopped. Expected results:

image

### DIFF
--- a/backend/npdfhir/middleware.py
+++ b/backend/npdfhir/middleware.py
@@ -33,7 +33,6 @@ class HealthCheckMiddleware:
                 health_status.update({
                     'status': 'unhealthy',
                     'database': 'disconnected',
-                    'error': str(e)
                 })
                 return JsonResponse(health_status, status=502)
 
--- a/backend/npdfhir/tests.py
+++ b/backend/npdfhir/tests.py
@@ -5,7 +5,6 @@ from fhir.resources.R4B.bundle import Bundle
 from pydantic import ValidationError
 from rest_framework import status
 from rest_framework.test import APIClient, APITestCase
-import json
 
 # I can't explain why, but we need to import cacheData here. I think we can remove this once we move to the docker db setup
 from .cache import cacheData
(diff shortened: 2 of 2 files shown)


backend/npdfhir/middleware.py
```diff
@@ -7,5 +12,29 @@ def __init__(self, get_response):
...
+
+            try:
+                # Test database connection
+                with connection.cursor() as cursor:
+                    cursor.execute("SELECT 1")
+                    cursor.fetchone()
+
+                return JsonResponse(health_status, status=200)
+
+            except Exception as e:
```
abachman-dsac: Given the lack of control on the type of `e`, it would be safer not to return the result of `str(e)` to the user making the healthcheck call.

We can log it still (putting off the discussion of structured logging for now), but outside of development environments we should keep responses limited.

backend/npdfhir/tests.py
```diff
@@ -160,8 +161,9 @@ class BasicViewsTestCase(APITestCase):
     def test_health_view(self):
         url = reverse("healthCheck")  # maps to "/healthCheck"
         response = self.client.get(url)
+        res_obj = json.loads(response.content.decode())
```
abachman-dsac: it would be cleaner to [use the built-in](https://docs.djangoproject.com/en/5.2/topics/testing/tools/#django.test.Response.json) `response.json()` here, which does the same thing.

```suggestion
        res_obj = response.json()
```

abachman-dsac (changes requested):
small change requested on the error response content and a tip (not blocker) on the test updates.

backend/npdfhir/middleware.py
tspecht-cms: removed error message from response

backend/npdfhir/tests.py
tspecht-cms: made this change

PR #130, opened 2025-10-23
Remove Unnecessary Columns

spopelka-dsac:
## Problem

There were some extraneous fields in the db that aren't relevant for the api or mapping

## Solution

This creates a new flyway migration file to remove those fields and updates the sample data insert statements accordingly.

## Result

Those extra columns will no longer be present in the application db after merge.

## Test Plan

1. navigate to `/backend`
2. run `make setup`; the db setup should run without error
3. run `make test`; the tests should run without error


dmtrscms:
Executed test plan w/no errors. It looks good to me.

PR #131, opened 2025-10-23
Format and Publish Backend Test Results

IsaacMilarky:
## Problem

We currently do run Django tests on PRs however we do not format and collate that data into the checks section for convenience. We also do not have docs to describe the tests and how they run on PRs.

## Solution

Use an upstream GitHub action to format the already present xml into a GitHub check. Also write some docs.

I am using [this](https://github.com/EnricoMi/publish-unit-test-result-action) upstream GitHub action to publish the xml results as a GitHub check. I have it configured to comment the test results as a comment on the PR and only create additional comments if there are a differing amount of failures from the last time that tests were ran. 

## Test Plan

Test the workflow on this branch before merging.

### DIFF
--- a/.github/workflows/backend-test.yml
+++ b/.github/workflows/backend-test.yml
@@ -45,6 +45,6 @@ jobs:
         if: (!cancelled())
         with:
           check_name: "Backend Django Test Results"
-          comment_mode: changes
+          comment_mode: changes in failures
           files: |
             backend/artifacts/test-reports/*.xml


abachman-dsac:
> ## Test Results
> 57 tests   57 ✅  0s ⏱️ 10 suites   0 💤 10 files     0 ❌
> 
> Results for commit [942ada4](https://github.com/DSACMS/npd/commit/942ada411e0f9ed0c098ff31732220113fc04ea7).

cuuuute, i dig it

PR #134, opened 2025-10-23
Cleaning Up Repo Workflows

sachin-panayil:
## Problem
Our contributors workflow kept failing because of branch protections. We also did not have the `update-codejson.yml `causing our `code.json` file to have old data. 

## Solution
- Refactor the `contributors.yml` file so that we compare the current contributors found in `COMMUNITY.md` and the actual number of them via Github API to check if the `contributors-readme-action` needs to run in the first place. 
- Add a shields.io badge to track the count so we dont have to handle 2 separate jobs. 
- Add the `update-codejson.yml` file

## Result
- If the `COMMUNITY.md` is out of date and the workflow needs to run, it will create a PR no longer causing failing jobs. The workflow will track if the file needs to be updated and will run based on that comparison. 
- Shields.io badge automatically tracks contributor count without any management
- Our code.json metadata file will now be updated on a monthly cadence keeping it fresh.

Downside is that we will have PRs that update contributors but they should be coming at a much lower rate. Its not possible to directly push to main with branch protections without some hacky workarounds so I think this is the best way to handle this

## Test Plan
- Run both jobs and check outputs
- Example of job running on personal fork [here](https://github.com/sachin-panayil/npd/actions/runs/18788131512/job/53611523369)
- Example of PR [here](https://github.com/sachin-panayil/npd/pull/1)

### DIFF
--- a/COMMUNITY.md
+++ b/COMMUNITY.md
@@ -51,7 +51,7 @@ See [CONTRIBUTING.md](CONTRIBUTING.md) for more details on the release process.
 
 ## Contributors
 
-![](https://img.shields.io/github/contributors/DSACMS/npd?style=flat-square&label=Contributors)
+![](https://img.shields.io/github/contributors/DSACMS/npd?style=flat-square&label=Contributor%20Count(incl.%20bots)) 
 
 <!-- readme: contributors -start -->
 <table>


abachman-dsac:
What is the effect of the new `update-codejson.yml` workflow?

sachin-panayil:
> What is the effect of the new `update-codejson.yml` workflow?

@abachman-dsac fields within code.json such as laborHours, dateModified, etc will be automatically updated once every month ensuring we have an up to date metadata file

PR #135, opened 2025-10-23
Wbprice/dagster ecs service attempt 2

wbprice:
## module-name: Dagster Definitions

image

## Problem

We don't have Dagster in the nonprod environment

## Solution

Defines Dagster infrastructure templates for dagster website and dagster daemon

## Result

Dagster exists in nonprod account

## Test Plan

Give your IP to @wbprice, then get the dagster ui url from @wbprice


dturner5234 (approved):
just needs files in s3 but it looks great

PR #136, opened 2025-10-24
Fix Documentation URLs

spopelka-dsac:
## Problem

We were only serving the swagger interface and not the redoc interface (which I think will be useful for the front end) nor the json representation of the docs. 

## Solution

This PR fixes the url patterns so that the swagger docs, redoc docs, and json docs can all be accessed. It also adds in a test for each.

## Result

Navigating to `/fhir/docs.json` returns JSON docs:
image

Navigating to `/fhir/docs/redoc` returns redoc docs:
image

Navigating to `/fhir/docs/*` returns swagger docs:
image

## Test Plan

1. Checkout `sjp/fix-doc-url`
2. navigate to `/backend`
3. run `make test`; all tests should pass
4. navigate to the above specified URLs. You should see the screenshotted docs

### DIFF
--- a/backend/npdfhir/tests.py
+++ b/backend/npdfhir/tests.py
@@ -28,6 +28,33 @@ def get_female_npis(npi_list):
     return results
 
 
+class DocumentationViewSetTestCase(APITestCase):
+    def setUp(self):
+        self.client = APIClient()
+
+    def test_get_swagger_docs(self):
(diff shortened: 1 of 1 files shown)


sachin-panayil (approved):
ran and LGTM. nice fix, this unlocks a level of automation within CapabilityStatement!

PR #137, opened 2025-10-24
Wbprice/simplify subnet setup

wbprice:
## Problem

The existing infrastructure splits the private subnets into "database subnets" and "etl subnets". I don't think there's any real advantage to doing this and it actually limits our options for high availability databases, etl container services in the future.

## Solution

Remove "db subnet" and "etl subnet" as concepts in the terraform code, specify "private subnets" instead

## Test Plan

Deploy dev infrastructure


dturner5234 (approved):
perfect

PR #140, opened 2025-10-24
Wbprice/run fhir api migrations against schema in etl database

wbprice:
## Problem

We need to manage the npd schema in the etl database for etl to fhir api mappings to work

## Solution

- Add flyway container to etl service, but limit migrations to the schema definitions and reference data (no fixture data) 

## Result

npd schema should be created in etl database without fixture data

## Test Plan

deploy to dev, inspect npd_etl database

### DIFF
--- a/infrastructure/modules/etl/main.tf
+++ b/infrastructure/modules/etl/main.tf
@@ -159,6 +159,10 @@ resource "aws_ecs_task_definition" "dagster_daemon" {
       ],
       secrets = [
         {
+          name      = "FLYWAY_PLACEHOLDERS_apiSchema"
+          value     = "npd_gold"
+        },
+        {
           name      = "FLYWAY_USER"
           valueFrom = "${var.db.db_instance_master_user_secret_arn}:username::"
         },
--- a/infrastructure/modules/fhir-api/main.tf
+++ b/infrastructure/modules/fhir-api/main.tf
@@ -127,6 +127,10 @@ resource "aws_ecs_task_definition" "app" {
       ],
       secrets = [
         {
+          name      = "FLYWAY_PLACEHOLDERS_apiSchema"
+          value     = var.db.db_instance_name
+        },
+        {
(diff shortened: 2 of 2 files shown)


infrastructure/modules/etl/main.tf
```diff
@@ -135,6 +140,42 @@ resource "aws_ecs_task_definition" "dagster_daemon" {
...
+    # where we have (for example) 4 API containers and 4 flyway containers and the 4 flyway containers all try to update
+    # the database at once. Flyway looks like it uses a Postgres advisory lock to solve this
+    # (https://documentation.red-gate.com/fd/flyway-postgresql-transactional-lock-setting-277579114.html).
+    # If we have problems, we can pull this container definition into it's own task and schedule it to run before new
+    # API containers are deployed
+    {
+      name      = "${var.account_name}-fhir-api-migration"
+      image     = var.fhir_api_migration_image
+      essential = false
+      command   = [ "migrate", "-locations=filesystem:/sql/migrations" ]
```
spopelka-dsac: We'll want to add /sql/reference_data here after merging https://github.com/DSACMS/npd/pull/139

wbprice: Rather than doing that, I've decided to only pack migrations and reference_data directories into the flyway images, that way the API migration reflects this change too

spopelka-dsac: nice!

infrastructure/modules/fhir-api/main.tf
```diff
@@ -126,6 +126,10 @@ resource "aws_ecs_task_definition" "app" {
         }
       ],
       secrets = [
+        {
+          name      = "FLYWAY_PLACEHOLDERS_apiSchema"
+          value     = var.db.db_instance_name
```
spopelka-dsac: just to confirm... this value resolves to npd?

PR #142, opened 2025-10-24
internal application load balancers

wbprice:
## Problem

- We want to restrict who has access to dagster ui (and the FHIR api, at lower envs)
- also, fix how apiSchema env var is specified in ECS

## Solution

configure terraform to create internal application load balancers, which are resolvable over zscaler.  fhir api visibility can be toggled on in prod, but lower envs are internal by default

## Test Plan

deploy envs/dev, note updated terraform outputs for API and dagster load balancers.


spopelka-dsac (commented):
Quick comment, otherwise lgtm

PR #144, opened 2025-10-26
Fix flyway migrations

dturner5234:
flyway migrations break when using custom flyway table names, overwrites the V1 migration...
This change updates the first migration to run as 1.1 to not conflict with flyways baseline migration


ftrotter-gov (approved):
This looks right and it is clearly needed to get these migrations to function on the ETL server as well as the NPD.

PR #148, opened 2025-10-27
Make npd.organization.authorized_official_id nullable

spopelka-dsac:
## Problem

Authorized official was not nullable, which would have created problems when inserting organizations that do not have an authorized official.

## Solution

This PR creates a new migration that drops the not null constraint from the authorized_official_id column in the npd.organization table, updates the views and serializers accordingly, and adds an associated test.

## Test Plan

1. navigate to `/backend`
2. run `make test`
3. The tests should pass without issue

If you're feeling wild, you can add your own organization to your local db, start up the API, and you should see the org come up, even if it doesn't have an authorized_official, like: 
image

### DIFF
--- a/.github/workflows/contributors.yml
+++ b/.github/workflows/contributors.yml
@@ -2,9 +2,6 @@ name: Update Contributors Information
 
 on:
   workflow_dispatch: {}
-  schedule:
-    # Weekly on Saturdays.
-    - cron: "30 1 * * 6"
   push:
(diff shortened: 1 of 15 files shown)


ftrotter-gov (approved):
This is the DB change that I need...

dmtrscms (approved):
Ran test plan. All tests passed. Good to go!

PR #157, opened 2025-10-28
increase database class for etl database to t3.large

wbprice:
## Problem

ETL database was crashing due to resource usage

## Solution

Increase the size of the etl database from t3.micro to t3.large

## Result

ETL database allocated RAM increases from 1GB to 8GB

## Test Plan

Review state of ETL database instance in RDS


spopelka-dsac (approved):
beefy

PR #169, opened 2025-10-30
update api db storage type

spopelka-dsac:
## Problem

We ran out of burst credits while transferring the etl data into the api db.
image

## Solution

This pr changes the storage type from gp2 to gp3 for the api db instance, so we don't have to reply on the burst credits.

Note: I already made this change in AWS so terraform plan should show no rds storage_type change for the api db instance


wbprice (approved):
terraform plan shows changes with the dagster services but not the database, so this is fine.

PR #176, opened 2025-11-03
Terraform files for DMS config

tspecht-cms:
## Problem

We need to use IaC using Terraform to create DMS resources for migrating an ETL database schema to the FHIR API database

## Solution

Do what the problem says to do.

## Result

All DMS resources are created and configured according to the environment

## Test Plan

TBD

### DIFF
--- a/infrastructure/modules/migrations/main.tf
+++ b/infrastructure/modules/migrations/main.tf
@@ -33,7 +33,7 @@ locals {
           "schema-name" : "npd",
           "table-name" : "%"
         },
-        "value" : "${var.etl_db.database_instance_name}-migrated"
+        "value" : "${var.etl_db.db_instance_name}-migrated"
       }
     ]
   })
--- a/infrastructure/envs/dev/main.tf
+++ b/infrastructure/envs/dev/main.tf
@@ -168,7 +168,7 @@ module "migrations" {
   source      = "../../modules/migrations"
 
   multi_az = false
-  account_name = var.account_name
+  account_name = local.account_name
   region = var.region
   tier = var.tier
   fhir_db = {


infrastructure/modules/migration/envs/dev/configs/table_mappings.json
```diff
@@ -0,0 +1,27 @@
...
+            "rule-id": 733828293,
+            "rule-name": "mhdlhs8h-od97g1",
+            "rule-type": "transformation",
+            "rule-target": "table",
+            "rule-action": "rename",
+            "object-locator": {
+                "schema-name": "npd",
+                "table-name": "%"
+            },
+            "value": "npd-dms-test"
```
wbprice: We probably would not use this specific table name, but we continue to use something other than "npd" in the short term.

Terraform has support for interpolating placeholder values in JSON, which we can use to configure this value at the module level: https://developer.hashicorp.com/terraform/language/functions/templatefile

infrastructure/modules/migration/envs/dev/configs/task_settings.json
```diff
@@ -0,0 +1,205 @@
...
+                "Severity": "LOGGER_SEVERITY_DEFAULT",
+                "Id": "FILE_TRANSFER"
+            }
+        ],
+        "LogConfiguration": {
+            "TraceOnErrorMb": 10,
+            "EnableTraceOnError": false
+        },
+        "CloudWatchLogGroup": "dms-serverless-replication-57J6Z4LH2JAUNKC3LS7RUETZUE",
+        "CloudWatchLogStream": "dms-serverless-serv-res-id-1761840201151-lqz"
```
wbprice: As above, we'd probably create log groups / streams in a Terraform context, then pass the name into this JSON config using template interpolation.

infrastructure/modules/migration/envs/dev/main.tf
```diff
@@ -0,0 +1,58 @@
...
+    source = {
+      database_name               = "npd_etl"
+      endpoint_id                 = "npd-east-dev-etl-db"
+      endpoint_type               = "source"
+      engine_name                 = "postgresql"
+      username                    = "npd_etl"
+      password                    = "youShouldPickABetterPassword123!" # where to get this?
+      port                        = 5432
+      server_name                 = "npd-east-dev-etl-db.cc3asw0omsvl.us-east-1.rds.amazonaws.com" # can this be a reference instead of hard coded?
+      ssl_mode                    = "require"
```
wbprice: Reference the `variables.tf` under the `fhir-api` or `etl` directory modules, which describe how to pass database information into a module.

The database information includes the database password ARN, which Terraform can use to fetch a secret value.

```hcl
# Fetch the current secret value
data "aws_secretsmanager_secret_version" "db" {
  secret_id = data.aws_secretsmanager_secret.db.id
}

# Decode the JSON secret string
locals {
  db_credentials = jsondecode(data.aws_secretsmanager_secret_version.db.secret_string)
}
```

infrastructure/modules/migration/envs/dev/main.tf
```diff
@@ -0,0 +1,58 @@
...
+      endpoint_id   = "npd-east-dev-fhir-api-db"
+      endpoint_type = "target"
+      engine_name   = "postgresql"
+      username      = "npd"
+      password      = "passwordsDoNotNeedToMatch789?" # where to get this?
+      port          = 5432
+      server_name   = "npd-east-dev-fhir-api-db.cc3asw0omsvl.us-east-1.rds.amazonaws.com" # can this be a reference instead of hard coded?
+      ssl_mode      = "require"
+    }
+  }
```
wbprice: because we're connecting to two databases, you'll need to update `variables.tf` to accept information about both databases.

infrastructure/modules/migrations/main.tf
```diff
@@ -0,0 +1,58 @@
...
+      migration_type            = "cdc"
+      replication_task_settings = file("configs/task_settings.json")
+      table_mappings            = file("configs/table_mappings.json")
+      source_endpoint_key       = "npd-east-dev-etl-db"
+      target_endpoint_key       = "npd-east-dev-fhir-api-db"
+
+      serverless_config = {
+        max_capacity_units     = 8
+        min_capacity_units     = 4
+        multi_az               = false # should be true for prod?
```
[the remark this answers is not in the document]
tspecht-cms: I created a new variable for this, set to `false` in Dev and `true` in Prod

infrastructure/envs/dev/variables.tf
```diff
@@ -6,6 +6,10 @@ variable "tier" {
   default = "dev"
 }
 
+variable "multi_az" {
+  default = false
+}
```
wbprice: Define this variable under `modules/migrations/variables.tf` instead of here, because it's specific to the migration module.

infrastructure/envs/prod/main.tf
```diff
@@ -163,6 +163,29 @@ module "etl" {
   }
 }
 
+module "migrations" {
+  source      = "../../modules/migrations"
```
wbprice: add your multi-az param here (true for prod)

infrastructure/modules/migrations/main.tf
```diff
@@ -0,0 +1,58 @@
...
+    source = {
+      database_name               = "npd_etl"
+      endpoint_id                 = "npd-east-dev-etl-db"
+      endpoint_type               = "source"
+      engine_name                 = "postgresql"
+      username                    = "npd_etl"
+      password                    = "youShouldPickABetterPassword123!" # where to get this?
+      port                        = 5432
+      server_name                 = "npd-east-dev-etl-db.cc3asw0omsvl.us-east-1.rds.amazonaws.com" # can this be a reference instead of hard coded?
+      ssl_mode                    = "require"
```
wbprice: It will, because even if the value is a string you need to be able to access a specific property within it

infrastructure/modules/migrations/main.tf
```diff
@@ -0,0 +1,108 @@
...
+
+  # Subnet group
+  repl_subnet_group_name        = module.networking.private_subnet_group_name
+  repl_subnet_group_description = "The DMS ETL replication subnet group"
+  repl_subnet_group_subnet_ids  = module.networking.private_subnet_ids
+
+  endpoints = {
+    etl-source = {
+      database_name = "npd_etl" # possible to reference this by env?
+      endpoint_id   = "npd-east-${var.tier}-etl-source"
```
wbprice: ${var.account_name} is a combination of `npd-east-` and `${var.tier}`, so you can use that instead of piping in tier directly.

infrastructure/modules/migrations/main.tf
```diff
@@ -0,0 +1,108 @@
...
+      server_name   = var.etl_db.db_instance_name # host name of the server - the correct value is npd-east-dev-etl-db.cc3asw0omsvl.us-east-1.rds.amazonaws.com
+      ssl_mode      = "require"
+    }
+
+    fhir-api-destination = {
+      database_name = "npd" # possible to reference this by env?
+      endpoint_id   = "npd-east-${var.tier}-fhir-api-destination"
+      endpoint_type = "target"
+      engine_name   = "postgresql"
+      username      = "npd" # possible to reference this by env?
```
wbprice: Yes, this is the `db_instance_username` output of the RDS module, but you'll need to update your `modules/migrations/variables.tf`, `envs/dev/main.tf`, `envs/prod/main.tf` to accommodate this.

infrastructure/modules/migrations/main.tf
```diff
@@ -0,0 +1,108 @@
...
+      engine_name   = "postgresql"
+      username      = "npd_etl" # possible to reference this by env?
+      password      = local.etl_db_password
+      port          = 5432
+      server_name   = var.etl_db.db_instance_name # host name of the server - the correct value is npd-east-dev-etl-db.cc3asw0omsvl.us-east-1.rds.amazonaws.com
+      ssl_mode      = "require"
+    }
+
+    fhir-api-destination = {
+      database_name = "npd" # possible to reference this by env?
```
wbprice: You want `var.fhir_db.db_instance_name`

infrastructure/modules/migrations/main.tf
```diff
@@ -0,0 +1,108 @@
...
+  endpoints = {
+    etl-source = {
+      database_name = "npd_etl" # possible to reference this by env?
+      endpoint_id   = "npd-east-${var.tier}-etl-source"
+      endpoint_type = "source"
+      engine_name   = "postgresql"
+      username      = "npd_etl" # possible to reference this by env?
+      password      = local.etl_db_password
+      port          = 5432
+      server_name   = var.etl_db.db_instance_name # host name of the server - the correct value is npd-east-dev-etl-db.cc3asw0omsvl.us-east-1.rds.amazonaws.com
```
wbprice: var.etl-db.db_instance_address

infrastructure/envs/prod/main.tf
```diff
@@ -163,6 +163,34 @@ module "etl" {
...
+
+  multi_az = true
+  account_name = local.account_name
```
wbprice: var

wbprice (approved):
`terraform -chdir=envs/dev plan` didn't return any errors. We can get this deployed to dev and validate more from there.

PR #177, opened 2025-11-03
[NDH-320] - Refactoring Search within Views

sachin-panayil:
## Problem

We now have three separate places where we update information about our endpoint search parameters. 

1. The ViewSet Logic
2. Swagger Documentation
3. CapabilityStatement 

This is three separate places that must be updated when a new search parameter is added or even a new endpoint. Lowering this burden would help a ton.

We should figure out a way to dynamically update all three of these at the same time. An idea I had was to create a central file that holds al our filters and then the other files can pull data from this central file. This means we would only need to update our search params in one place and the changes can propagate throughout the rest of the code

## Solution

Use `django-filters` to fulfill 1  by creating a custom backend that our views can tap into

## Result

Search has been broken out of the views and we only need to edit the `filters.py` file to make updates on all three areas

* Swagger Documentation will be broken out into a separate PR
* CapabilityStatement will also be included since we need to use a different route for Swagger PR
* The idea is that filters.py will be the central source of truth for the three search parameter locations

## Test Plan

- run `docker compose up`
- test search for each endpoint


backend/npdfhir/filters.py
abachman-dsac: Let's get ahead of file creep and move this to a folder structure: 

```
filters/
  __init__.py
  endpoint_filter_set.py
  practitioner_filter_set.py
  ...
```

etc.

backend/npdfhir/utils.py
```diff
@@ -24,4 +24,15 @@ def genReference(url_name, identifier, request):
         reverse(url_name, kwargs={'pk': identifier}))
     reference = Reference(
         reference=reference)
-    return reference
\ No newline at end of file
+    return reference
+
+def parse_identifier(identifier_value):
```
abachman-dsac: nit: this method name is _very_ generic. 

i know you're just moving it up a level, organizationally, though.

abachman-dsac (approved):
a clean split of the filters, thanks for taking care of that!

looks like a good refactor. 

we'll want to eventually introduce tests that are more specific about the data which is retrieved / retrievable by each filter, but this gives us a place to put them

PR #178, opened 2025-11-03
[NDH-343] flyway migration to bootstrap superuser in deployment environments

abachman-dsac:
## Problem

We need to run `python manage.py createsuperuser` to create our first superuser for the system, but Infrastructure-as-Code dictates that we shall not run one-off commands by SSH-ing into a box and typing `python` at the console :P

## Solution

Added a repeatable flyway reference data migration which will create a superuser account when migrations happen during deployment.

## Result

Developers can ignore this migration, since the `make createsuper` command already exists.

Once the migration has been run in all relevant non-ETL deployment contexts and an initial superuser account exists, that superuser can create new user accounts and the migration, can be removed from the project.

## Test Plan

We will need to verify that this terraform plan works in dev before attempting a deployment to production.

### With external data source

This PR introduces a few new Terraform objects:

- AWS SecretsManager data source to generate a random password
- external data source which can provide a PBKDF2 hashed password string 
    - supported by `infrastructure/modules/fhir-api/generate_hash.py`, a python script which does the work
- AWS SecretsManager resource to store the original password and the hashed value

The stored `hashed_password` value is later used by the FHIR API migration task during deployment as the value of the `FLYWAY_PLACEHOLDERS_superuserDefaultPassword` environment variable. 

Thus, after the first successful deployment with this setup, a new user account with the username `npd+deploy@cms.hhs.gov` will exist and its password will be stored in AWS SecretsManager under the secret with prefix `${var.account_name}-fhir-api-superuser-default-password`.

### Without external data resource

This PR requires a secret to be stored in AWS, generated by terraform. The content of the secret will be a key / value object with the keys `password` and `hashed_password`. 

After the first deployment, the new Secrets Manager resource `password` key will contain an auto-generated password and `hashed_password` will be an empty string. Given the logic expressed in the repeatable flyway migration included in this PR, no superuser account will be created.

An administrator with read/write access to the appropriate AWS cloud environment will be able to follow-up and use the instructions described below to generate a valid PBKDF2 password hash for the password generated in AWS by the terraform configuration described in this PR.

Once generated, the AWS Secrets Manager resource should be updated in the `hashed_password` field so that the next deployment will trigger the repeatable flyway migration to actually create the superuser record. 

## Developer's guide to generating a PBKDF2 string

<details>

<summary>To generate a PBKDF2 string, you can follow the steps described here.</summary>

Create superuser account locally: 
```sh
docker compose run --rm \
	-e DJANGO_SUPERUSER_EMAIL="npd@cms.hhs.gov" \
	-e DJANGO_SUPERUSER_USERNAME="npd@cms.hhs.gov" \
	-e DJANGO_SUPERUSER_PASSWORD="lNZnMXx7_FjeSN_OOwBHzmgkfylNJtcT" \
	django-web python manage.py createsuperuser --no-input
```

Export the raw `password` column value:
```sh
docker compose run --rm \
	 python manage.py shell \
	 -c "print(User.objects.get(username='npd@cms.hhs.gov').password)" 

# => pbkdf2_sha256$1000000$245jUUMjNMsyXEy39nDMIZ$VfbNC9viFFFQ2yuZq6SXeB90e5F42p+tpjtjRWWD0uk=
```

You can run the migration included in this PR by passing `superuserDefaultPassword` as a flyway placeholder to the `migrate` command: 

```sh
# by env var
docker compose run --rm \
	-e FLYWAY_PLACEHOLDERS_superuserDefaultPassword='pbkdf2_sha256$1000000$...' \
	db-migrations migrate

# by CLI argument
docker compose run --rm \
	db-migrations migrate -placeholders.superuserDefaultPassword='pbkdf2_sha256$1000000$...'
```

**NOTE**: the PBKDF2 formatted password includes `$` symbols, and so must be single quoted in order to execute successfully from a shell command.

On subsequent runs of the `migrate` command, the default value of `FLYWAY_PLACEHOLDERS_superuserDefaultPassword` in `backend/docker-compose.yml` is set to `""` (empty string) and so the repeatable migration will run, but the `password` value of the superuser account will not change.

If you want to update the superuser account password from a future migration, repeat the set + export steps described above and pass the new value of `password` to a flyway migration command as `FLYWAY_PLACEHOLDERS_superuserDefaultPassword`. Re-runnable migrations will be automatically executed _EVERY TIME_ the script changes, and a changed value for a placeholder counts.

</details>

### DIFF
--- a/infrastructure/modules/etl/main.tf
+++ b/infrastructure/modules/etl/main.tf
@@ -200,10 +200,6 @@ resource "aws_ecs_task_definition" "dagster_daemon" {
         {
           name = "FLYWAY_BASELINE_ON_MIGRATE"
           value = "true"
-        },
-        {
-          name  = "FLYWAY_PLACEHOLDERS_superuserDefaultPassword"
-          value = ""
         }
       ]
       secrets = [


infrastructure/modules/fhir-api/main.tf
```diff
@@ -51,6 +51,23 @@ resource "aws_iam_policy" "fhir_api_can_access_fhir_api_db_secret" {
...
+      {
+        Action = "secretsmanager:GetSecretValue",
+        Effect = "Allow"
+        Resource = [
+          var.fhir_api_superuser_password_arn
+        ]
+      }
+    ]
+  })
+}
```
wbprice: This needs a corresponding role/policy attachment, which associates the policy to the execution role.

```hcl
resource "aws_iam_role_policy_attachment" "fhir_api_can_access_fhir_api_superuser_secret_attachment" {
  role       = aws_iam_role.fhir_api_execution_role.name
  policy_arn = aws_iam_policy.fhir_api_can_access_fhir_api_superuser_password.arn
}
```

infrastructure/modules/etl/main.tf
```diff
@@ -200,6 +200,10 @@ resource "aws_ecs_task_definition" "dagster_daemon" {
         {
           name = "FLYWAY_BASELINE_ON_MIGRATE"
           value = "true"
+        },
+        {
+          name  = "FLYWAY_PLACEHOLDERS_superuserDefaultPassword"
+          value = ""
```
wbprice: This ought to be under `secrets` if you want to read from a secret.

flyway/sql/reference_data/R__1_create_superuser.sql
```diff
@@ -0,0 +1,33 @@
...
+        username,
+        email,
+        first_name,
+        last_name,
+        is_staff,
+        is_active,
+        date_joined
+    )
+SELECT
+    '${superuserDefaultPassword}',
```
wbprice: I'm assuming this comes from `FLYWAY_PLACEHOLDERS_superuserDefaultPassword` but `FLYWAY_PLACEHOLDERS_` is stripped from the key?

infrastructure/modules/etl/main.tf
wbprice: Or is the intention to make this blank on purpose, the ETL migration scripts don't need this?

abachman-dsac: yeah, this is blank on purpose because ETL doesn't need it. 

now that you mention it, though, ETL never runs the django flyway migrations, so it will probably blow up because `public.auth_user` won't exist in the ETL DB.

infrastructure/modules/fhir-api/main.tf
```diff
@@ -51,6 +51,23 @@ resource "aws_iam_policy" "fhir_api_can_access_fhir_api_db_secret" {
...
+  name        = "${var.account_name}-fhir-api-can-access-fhir-superuser-password"
+  description = "Allows ECS to access the pbkdf2 encoded default superuser password"
+  policy = jsonencode({
+    Version = "2012-10-17"
+    Statement = [
+      {
+        Action = "secretsmanager:GetSecretValue",
+        Effect = "Allow"
+        Resource = [
+          var.fhir_api_superuser_password_arn
```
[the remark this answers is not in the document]
abachman-dsac: tl;dr - i found a terraform-y way to implement a generator which can create a hashed password based on an `aws_secretsmanager_random_password` data source random password.

infrastructure/modules/etl/main.tf
wbprice: I'm hearing "the ETL migration doesn't need this variable" so let's remove it.

wbprice (approved):

PR #179, opened 2025-11-03
bump etl database storage

wbprice:
ETL database is full and refusing connections. Bumping the provisioned storage to 250 GB while we figure out what is taking up the space.


dturner5234 (approved):
I was seeing 100GB usage locally. 200 should be enough. we might be able to refactor some of these tables into views and reduce this a bit eventually.

PR #186, opened 2025-11-06
Update security groups for application load balancers to accept port …

wbprice:
…80 and 443 connections

## Problem

Various team members and stakeholders can't access the internal routes. It could very well be a Zscaler thing but updating the security groups to not check against the prefix list or cmscloud vpn fixes this in the short-term with minimal risk (all the relevant endpoints are internal anyway)

## Solution

Update security group configuration for FHIR API and Dagster UI internal load balancers.

## Result

Internal users can access the internal routes but external users can't

## Test Plan

deploy and check the routes


abachman-dsac (approved):
I am for this approach.

DNS will, of course, become more complex in the future, but this gets us over our first hump: can everyone on the CMS team reach the non-production site?

PR #188, opened 2025-11-06
update pglogical configuration pending reboot

wbprice:
This RDS parameter needs to be set pending-reboot. This was causing production infrastructure to not deploy.


dturner5234 (approved):
perfect

PR #194, opened 2025-11-07
[NDH-340] Porting to drf-spectacular

sachin-panayil:
## Problem
Currently, we are using a old library to support our Swagger Documentation, drf-yasg. This library also only supports OpenAPI 2.0. This library doesnt intergrate with django-filters and is old

## Solution
By upgrading to drf-spectacular, we gain the ability to integrate with OpenAPI 3.0 and some other cool abilities that help with auto-generation fo documentation

## Result
We gain the ability to automatically generate in depth documentation such as our views search parameters. We also are now using OpenAPI 3.0.

* Response examples is something that needs to be discussed further. That was the goal for the original ticket but this PR can get out while we work on response examples moving forward.

## Test Plan
- run docker compose up 
- check the swagger documentation /fhir/docs

### DIFF
--- a/backend/npdfhir/views.py
+++ b/backend/npdfhir/views.py
@@ -62,7 +62,6 @@ class FHIREndpointViewSet(viewsets.GenericViewSet):
     @extend_schema(
         responses={
             200: OpenApiResponse(
-                response=BundleSerializer,
                 description='Successfully retrieved FHIR Bundle resource of FHIR Endpoint resources'
             )
         }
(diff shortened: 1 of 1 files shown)


spopelka-dsac (commented):
The filter fields look awesome!! It would be great to include defaults and max values for the page size, if possible (reading dynamically from those variables).

I'm noticing a few discrepancies with the sample responses (e.g. 

Image

vs. 

Image

I'd really like to land the filter fields and updated OpenAPI version, since ACV will be reviewing the documentation, so maybe we move the schemas to another PR and continue to work that separately?

PR #205, opened 2025-11-12
Set NPD_REQUIRE_AUTHENTICATION in dev and prod through terraform

abachman-dsac:
## Problem

We need to be able to control our deployed environments' authentication requirement.

For launch, we would like nonprod to _not_ (yet) require authentication and production to require it.

## Solution

Make sure `NPD_REQUIRE_AUTHENTICATION` is set appropriately in the fhir-api containers for Dev and Prod.

## Result

Dev will continue to not require authentication, while Prod will require it from the beginning.


wbprice (commented):
Shouldn't we test that the authentication works in lower environments before turning it on in prod?

PR #218, opened 2025-11-18
ensure header title does not wrap at desktop size

abachman-dsac:
## Problem

New responsive design of header breaks at desktop resolution.

_before_: 

image

## Solution

Add `whitespace: nowrap` to the text component of the header logo element.

## Result

_after_:

Desktop:

image

---

Tablet: 

image

---

Mobile: 

image


sman-dsac (approved):
Looks good! Can you confirm nothing broke in the mobile or tablet width views?

PR #233, opened 2025-11-26
self-referencing rules for etl and fhir api db security groups

tspecht-cms:


infrastructure/modules/migrations/main.tf
```diff
@@ -90,9 +99,17 @@ module "database_migration_service" {
       serverless_config = {
         max_capacity_units     = 8
         min_capacity_units     = 4
-        multi_az               = var.multi_az
-        vpc_security_group_ids = [var.networking.etl_db_security_group_id, var.networking.api_db_security_group_id]
+        multi_az               = local.multi_az
```
wbprice: Why did this change from var to local? Are you computing multi_az in a local block?

infrastructure/modules/migrations/main.tf
```diff
@@ -1,39 +1,48 @@
...
 data "aws_secretsmanager_secret_version" "fhir_db" {
   secret_id = var.fhir_db.db_instance_master_user_secret_arn
 }
 
 locals {
-  etl_db_password  = jsondecode(data.aws_secretsmanager_secret_version.etl_db.secret_string).password
-  fhir_db_password = jsondecode(data.aws_secretsmanager_secret_version.fhir_db.secret_string).password
-  account_name     = "${var.region}-${var.tier}"
-  multi_az         = var.multi_az
+  etl_db_password = try(
```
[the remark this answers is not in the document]
wbprice: wouldn't we want this to fail if a password couldn't be found?

infrastructure/modules/migrations/main.tf
tspecht-cms: This appears to be an oversight. I'll change it back.

PR #234, opened 2025-11-26
[NDH-448] Modify Django Tests to Use setUpTestData Methods Instead of Sample Data SQL

IsaacMilarky:
## Problem

Right now we are relying on pre-populated sample data defined in raw SQL to run our tests on proper test data. This is a problem because it makes it less convenient to test when trying to refer to uuids or have dynamic tests that are verified based on data that the test knows is supposed to be correct. 

## Solution

Make it so that the `compose-test.yml` does not insert the sample data at all. Instead use the setupTestData method in each set of tests to insert proper data that can be checked by the tests. 

## Result

Summary:
* Alter `compose-test.yml` to override the volumes section to ignore the `local_dev` directory
* Create new `fixtures.py` file to have methods that can create test records given the proper information
* Use those methods in the test files to create `setUpTestData` methods that insert test data

## Test Plan

Run `make test` and ensure that the tests are running properly

### DIFF
--- a/backend/npdfhir/tests/fixtures.py
+++ b/backend/npdfhir/tests/fixtures.py
@@ -144,9 +144,7 @@ def create_organization(
         if organization_type:
             code = Nucc.objects.get(pk=organization_type)
 
-            OrganizationToTaxonomy.objects.create(
-                npi=clinical_organization, nucc_code=code
-            )
+            OrganizationToTaxonomy.objects.create(npi=clinical_organization, nucc_code=code)
 
     OrganizationToName.objects.create(
         organization=org,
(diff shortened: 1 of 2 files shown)


abachman-dsac (changes requested):
Please make sure you're using the `ruff` formatter and linter on all code in your branch. 

The easiest way to do that would be:

- make sure your branch is up to date with `main`
- from the `backend/` directory:
    - run `make install-tools`
    - run `make format` 

in the backend folder.

sachin-panayil (approved):
lgtm. this was much needed and makes our testing way less brittle 🔥

PR #237, opened 2025-12-01
[NDH-424] Unified project tooling

abachman-dsac:
## developer tooling: Centralize make and docker compose, introduce bin/npr

[Jira Ticket NDH-424](https://jiraent.cms.gov/browse/NDH-424)

## Problem

As a developer: 

- the full-stack NPD development "happy path" still has too many things to remember
- `make` commands and docker compose services in `backend/` are at the wrong scope for what they describe
    - e.g., `make migrate` runs flyway commands in a container built from the `../flyway` directory
- working on a new project is easier when common workflows can be run from the top-level of the project

One goal of these changes is that they _SHOULD NOT_ require a completely stripped down rebuild of the project for any engineers who are already running it. 

## Solution

Updates:

- I moved the bulk of our existing `backend/Makefile` into the root of the project
- I moved all docker compose service definitions into the top-level `compose.yml` file
- I introduced the `bin/npr` script as a helper for running common language / service-specific commands inside the appropriate docker compose service containers
- related change: the frontend `web` service will run `npm run watch` instead of `npm run dev` to ensure that we're always doing development against static assets hosted at http://localhost:8000, by default

### Setup changes

`make setup && make install-tools` should now suffice to go from zero to ready system.

### Testing changes

`make test` will run the backend and frontend-specific test suites.

Run `make test-server` to start a playwright-specific test instance of the django-web service and in another terminal, `make playwright` to run the end-to-end test suite locally.

The backend and frontend suites can be run separately with `make test-backend` and `make test-frontend`, respectively after running `test-setup`

### Documentation changes

This PR includes updates to the CONTRIBUTING.md guide for developers and updates README.md to reduce the amount of superfluous documentation.

## Test Plan

Shut down all docker compose services if you're running them from inside the `backend/` directory. 

Try out the `make` commands you were using before from the project root to make sure they work as expected with your setup.

Read through the help documentation printed out by `make` and `bin/npr` to see what they do.

### DIFF
--- a/compose.test.yml
+++ b/compose.test.yml
@@ -24,7 +24,6 @@ services:
       - ./flyway/sql/django_reference_data:/flyway/sql/django_reference_data
       - ./flyway/sql/migrations:/flyway/sql/migrations
       - ./flyway/sql/reference_data:/flyway/sql/reference_data
-      - ./flyway/sql/local_dev:/flyway/sql/local_dev
 
   django-web:
     extends:


spopelka-dsac:
Also I'm realizing that the Django tests are failing...

PR #198, opened 2025-11-12
[NDH-398] Organization details screen - initial work

abachman-dsac:
## NPD app: Creates initial organization show page, styling and data connection

[Jira Ticket NDH-398](https://jiraent.cms.gov/browse/NDH-398)

## Problem

We need to show details about an Organization. 

We also need to establish a clean data access pattern for the frontend / NPD browser app.

## Solution

Added an Organization lookup / detail view page at `/organizations/:organizationId` which passes the given ID straight through to the FHIR API at `/fhir/Organization/{id}` to retrieve an Organization record.

I also introduced [@tanstack/react-query](https://tanstack.com/query/latest/docs/framework/react/overview) as a solution for making requests from the frontend and established a pattern for defining expected API response types.

## Result

If the `SEARCH_APP` feature flag is true on your system, you can browse to a known Organization URL by ID. 

If the `ORGANIZATION_LOOKUP_DETAILS` feature flag is true, you can see the full details page.

### Future work

The use of the `<FeatureFlag>` component to wrap incomplete display of information is intended to permit continuing work on the Organization view page without displaying placeholders to people outside the appropriate group permissions. In this case, the `Developers` group will be used to make sure that we are not presenting confusing or incomplete details

This PR **DOES NOT** attempt to populate the full details, that work will be part of a future ticket: [NDH-476](https://jiraent.cms.gov/browse/NDH-476).

### Screenshots

_with `SEARCH_APP` and `ORGANIZATION_LOOKUP_DETAILS` feature flags set_:

with-visibility

---

_with `SEARCH_APP` but not `ORGANIZATION_LOOKUP_DETAILS`_:

organization-view-without-features

### Automated testing

There are new unit tests in the frontend directory covering: 

- basic API access functionality
- use of feature flags
- the new Organization viewer

### User acceptance testing

Automated e2e testing for the actual API interactions are part of #227 and not yet merged, so for this PR you can do the following:

1.  update development environment
    * from scratch with `make setup` OR run migrations with `docker compose run --rm db-migrations migrate` 
    * update frontend dependencies with `docker compose run --rm web npm install`
    * MAKE SURE your local migrations have run  
3. make a request to `http://localhost:8000/fhir/Organization/` to get a valid organization ID, for example `"3ac907f0-c7d0-484e-ad0e-653dd011a55b"`
    ```console
    $ curl "http://localhost:8000/fhir/Organization/" | jq -r '.results.entry[0].resource.id'
[...]

PR #239, opened 2025-12-02
Automate export of FHIR API schema to Typescript

abachman-dsac:
## Problem

In the initial Organization details page PR, #198, I took a shortcut to generating a Typescript `interface` for FHIR Organization API responses, this approach falls apart quickly as I start work on the Practitioner details page.

We want to be able to derive Typescript interface definitions from Pydantic base models in as few steps as possible.

## Solution

Introduce a new Django management command for exporting FHIR schema exports and a javascript library for converting those exports to Typescript.

## Result

The new `frontend/src/@types/fhir` directory contains schemas, auto-generated types, and a README.md describing how to produce / re-produce these records.

Additionally, I updated the existing frontend code for accessing the Organization API to properly account for the updated type definitions.

Lastly, I deleted the old `@types/fhir_api.d.ts` type definition file, since it was incorrect.

## Test Plan

Automated test coverage is adequate for this change.

### DIFF
--- a/frontend/tests/fixtures/index.ts
+++ b/frontend/tests/fixtures/index.ts
@@ -1,3 +1,4 @@
+import type { Organization } from "../../src/@types/fhir/Organization"
 import fhirOrganization from "./fhir_organization.json"
 
 export const DEFAULT_FRONTEND_SETTINGS: FrontendSettings = {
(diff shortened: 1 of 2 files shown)


backend/npdfhir/management/commands/exportschema.py
```diff
@@ -0,0 +1,127 @@
...
+        parser.add_argument(
+            "--record",
+            help="only export the an example record JSON dump",
+            action="store_true",
+            default=False,
+        )
+        parser.add_argument(
+            "--username",
+            help="username of a local NPD user (required if authentication is required)",
+            default="developer@cms.hhs.gov",
```
sachin-panayil: do we want to have this out in public/default?

sachin-panayil (approved):
have once concern but considering that info lives in other places, should be good to go. this file will save a good amount of time! i wonder if its worth breaking this out into a seperate package/repo 👀

PR #242, opened 2025-12-03
[NDH-476]: Organization Detail MVP

sachin-panayil:
## Problem
After initial creation of the Organization Detail page, the rest of the page needed to be fleshed out with the current endpoints. This requires us to use actual data and to replace static text to use i18n

## Solution
Start using real data to populate the sections based on current limitations. Replace all static text with a new organizations.json for i18n

## Result
Organizations now fill up the About, Contact and Identifiers sections with real data. Static text account for all current and future translations

Screenshot 2025-12-04 at 5 04 14 PM

## Test Plan
- on root, run `docker compose up django-web`
- then run `docker compose run --rm web npm run watch`
- login in via http://localhost:8000/admin/login/
    - username: developer@cms.hhs.gov
    - password: password123
- go to /[organization/id](http://localhost:8000/organizations/ade8cd26-dcb5-4ed6-ae56-e2bdd46b0cc8)

### DIFF
--- a/frontend/src/state/requests/organizations.ts
+++ b/frontend/src/state/requests/organizations.ts
@@ -1,12 +1,11 @@
 import { useQuery } from "@tanstack/react-query"
-import type { Organization } from "../../@types/fhir/Organization"
 import { apiUrl } from "../api"
 import { formatAddress, formatDate } from "../../helpers/org_helpers"
 import type { FHIROrganization } from "../../@types/fhir"
 
 const fetchOrganization = async (
   organizationId: string,
-): Promise<Organization> => {
+): Promise<FHIROrganization> => {
   const url = apiUrl("/fhir/Organization/:organizationId/", { organizationId })
 
   const response = await fetch(url)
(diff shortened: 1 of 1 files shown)


frontend/src/i18n/en/organizations.json
```diff
@@ -0,0 +1,16 @@
+{
+    "title": "Provider Group",
+    "unavailable": "This content is not currently available.",
+    "update": {
+        "title": "Are you the practitioner listed?",
+        "subtitle" : "Learn how to ",
```
abachman-dsac: if we need whitespace in any particular context, let's do that in the React component where this template literal is referenced rather than as part of the string itself. 

Worth changing only because "has trailing whitespace" is a significant implementation detail, but is almost impossible to enforce with typescript or testing, whereas just putting `{' '}` into the React component is a hard guarantee.

A safe implicit assumption is "template strings may be trimmed before rendering" and so the component should always handle layout concerns like how things are spaced out.

abachman-dsac:
_not a change request, but noting for future work_

I like the idea of adopting `FHIR` as a prefix for all publicly exposed API response types.

Along with that, we should be careful to ensure that outside of `@types/fhir`, all code only ever references the types defined in `@types/fhir/index.ts`, which are under our supervision, rather than drilling into `@types/fhir/Thing.ts` where they would be subject to code generation tooling.

To think a few steps into the future, there is a use case for a standalone npm package for API response types provided by NPD. Writing the rest of the application in a way that respects that, therefore, is prudent.

PR #244, opened 2025-12-03
correct typo in lvextend instructions

wbprice:
The directions are currently recommending growing the root volume twice, which is not desired.


abachman-dsac (approved):
which approved

PR #249, opened 2025-12-04
[NDH-553] Basic React error pages

abachman-dsac:
[Jira Ticket NDH-553](https://jiraent.cms.gov/browse/NDH-553)

## Problem

Every React app has a "white screen of death" when errors go unhandled. 

Two errors every single page app with routing will face:

- router has no path for that URL
- everything else

## Solution

Added a simple React ErrorBoundary component and inserted it into the main.tsx hierarchy. 

Added a generic NotFound page and put it in the routing tree with the catch-all path.

## Result

invalid paths look like this:

localhost_8000_not_real

unhandled errors look like this:
 
localhost_8000_

## Testing

"not found" is easy, start up the app and visit http://localhost:8000/not/a/real/page. I also added a new end-to-end test for this.

Unhandled errors require writing intentionally invalid code. Easiest way I've found is to put a `throw` in a React component, for the screenshot `frontend/src/pages/Layout.tsx:7-11`:

```tsx
// A react-router compatible page wrapper.
export const Layout = () => {
  const { loading } = useFrontendSettings()

  throw new Error("oh no!")
```

### DIFF
--- a/frontend/src/components/ErrorBoundary.tsx
+++ b/frontend/src/components/ErrorBoundary.tsx
@@ -1,22 +1,55 @@
+import { Button } from "@cmsgov/design-system"
 import React from "react"
+import { Footer } from "./Footer"
+import { Header } from "./Header"
+
+import classNames from "classnames"
+import layout from "../pages/Layout.module.css"
 
 interface ErrorFallbackProps {
   error: Error
   resetErrorBoundary?: () => void
 }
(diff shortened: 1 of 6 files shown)


sachin-panayil (approved):
lgtm! i wonder if @sman-dsac has some ideas about how these pages should look

PR #250, opened 2025-12-05
[NDH-324] Improve Practitioner Test Assertions

IsaacMilarky:
## Problem

The test_practitioner.py file in the tests directory has tests that only test whether results are returned, not whether the name filter is properly filtering results.

## Solution

Add assertions to the tests to make sure the results of the test query are being properly filtered. For example, making sure that if you filter practitioners by US state using address filters the tests will make sure that every result has addresses that are located in that US state. 

## Result

Summary:

* Improve the fixtures.py logic to support Practitioners being associated with a Location in the tests
* Allow the fixtures.py logic to assign Nucc codes when provided 
* Fix an oversight in the create_location method that did not properly assign the US state to an address with a FipsState code
* Add to the practitioner fixture data additional data corresponding to Nucc classification and location. 
* Add assertions making sure that the results returned when querying a practitioner by name match the parameters
* Add assertions making sure that the practitioner_type filter fetches the proper record matching a Nucc code
* Add assertions making sure that the endpoint can properly filter by NPI by checking the results match the NPI parameter
* Add additional assertions making sure that the address filtering returns results that match the locations on the fixture data

## Test Plan

Run `make test`


abachman-dsac:
I am chill keeping all this as is, but want to note that we are repeating the same pattern of "one file to rule them all" with [fixtures.py](https://github.com/DSACMS/npd/blob/main/backend/npdfhir/tests/fixtures.py) and have started building another thing which should be a proper module, broken apart by domain / subject.

In this case, I would say the strongest smell is https://github.com/DSACMS/npd/blob/e70149a3c93119819b51d1a135f29f81efd758d3/backend/npdfhir/tests/fixtures.py#L4-L37, a single `import` statement that brings in 30+ other classes. 

The import is more maintainable now that we have linting and auto-formatting, but still creates a module that is difficult to read, maintain, and reason about.

PR #256, opened 2025-12-09
[NDH-324] Amendments to Fixtures and Test cases

IsaacMilarky:
## Problem

The current state of the amended test fixtures is not quite to specification for ticket NDH-324. Also, the fixtures file is getting to be very large and needs to be turned into a module of separate files. 

## Solution

Amend the test cases to check all returned data and make sure that we aren't getting back data that was not supposed to be fetched by the given filter parameter.  Also, improve the tests by adding additional fixture data that adds a non NPI identifier. Additionally, improve the test cases by making sure all assertions match the filters. 

I have also created a new fixtures module to replace the fixtures.py file because it was getting to be pretty large.

## Result

Summary:

* Amend test cases to check all results
* Amend fixture data to create a practitioner with a non NPI identifier
* Amend address test case to test for a more specific search parameter than "Street"
* Test that the address part of the returned data includes use data
* Amend test case to check the 'use' field instead of other identifying data
* Edit test data fixture to have multiple codes

The amended tests have also exposed some issues in the code-base. Namely, the test that checks to see if the identifier filter can fetch non NPI identifiers does not find that identifier in the returned data. I think this is because the non NPI identifier is just not present in the serialized data although the correct record does appear to be fetched.

Additionally, the tests have also found that the 'use' field is missing from the returned practitioner data address field. 

I also ran `make format` so there have been a bunch of formatting changes from ruff.

### DIFF
--- a/frontend/src/pages/Developers/Developers.content.md
+++ b/frontend/src/pages/Developers/Developers.content.md
@@ -10,7 +10,7 @@ Please note the following:
 - Frequent updates and API changes are possible
 - Feedback is encouraged
 
-#### Providing feedback
+## Providing feedback
 
 Feedback in all areas is welcome during the beta, including the following:
 


backend/npdfhir/tests/fixtures/endpoint.py
```diff
@@ -0,0 +1,77 @@
...
+    EndpointType,
+    EnvironmentType,
+    PayloadType,
+)
+from .organization import create_organization
+
+
+def _ensure_endpoint_base_types():
+    """
+    Flyway inserts some reference values, but ensure minimal ones exist.
```
spopelka-dsac: Can you say more about the behavior you were experiencing that necessitated these _ensure_ functions?

backend/npdfhir/tests/test_practitioner.py
```diff
@@ -222,21 +226,23 @@ def test_list_filter_by_name(self):
...
-        ][0]["code"]
-        self.assertEqual(self.nurse_code, nurse_code)
+        for entry in response.data["results"]["entry"]:
+            nurse_codes = [
+                nc["code"] for nc in entry["resource"]["qualification"][0]["code"]["coding"]
+            ]
+            self.assertIn(self.nurse_code, nurse_codes)
 
     # Identifiers Filter tests
     def test_list_filter_by_npi_general(self):
```
spopelka-dsac: I think it is clearer if this test is named test_list_filter_by_identifier_general

backend/npdfhir/tests/test_practitioner.py
```diff
@@ -245,28 +251,35 @@ def test_list_filter_by_npi_general(self):
...
         assert_has_results(self, response)
 
-        result_city_string = response.data["results"]["entry"][0]["resource"]["address"][0]["city"]
-        self.assertEqual(city_string, result_city_string)
+        for entry in response.data["results"]["entry"]:
+            result_city_string = entry["resource"]["address"][0]["city"]
+            self.assertEqual(city_string, result_city_string)
+            result_addr_string = entry["resource"]["address"][0]["line"][0]
+            # print(result_addr_string)
+            self.assertEqual(address_line, result_addr_string)
```
spopelka-dsac: The address search is a partial text match, so this should be assertIn not assertEqual

backend/npdfhir/tests/test_practitioner.py
```diff
@@ -245,28 +251,35 @@ def test_list_filter_by_npi_general(self):
...
+        address_line = self.locs[2].address.address_us.delivery_line_1
+        response = self.client.get(url, {"address": "123 Street R. Rochester"})
         self.assertEqual(response.status_code, status.HTTP_200_OK)
         assert_has_results(self, response)
 
-        result_city_string = response.data["results"]["entry"][0]["resource"]["address"][0]["city"]
-        self.assertEqual(city_string, result_city_string)
+        for entry in response.data["results"]["entry"]:
+            result_city_string = entry["resource"]["address"][0]["city"]
+            self.assertEqual(city_string, result_city_string)
```
spopelka-dsac: I'm not sure that we need this assertion here

backend/npdfhir/tests/test_practitioner.py
```diff
@@ -245,28 +251,35 @@ def test_list_filter_by_npi_general(self):
...
         city_string = self.locs[2].address.address_us.city_name
-        response = self.client.get(url, {"address": "Street"})
+        address_line = self.locs[2].address.address_us.delivery_line_1
+        response = self.client.get(url, {"address": "123 Street R. Rochester"})
         self.assertEqual(response.status_code, status.HTTP_200_OK)
         assert_has_results(self, response)
 
-        result_city_string = response.data["results"]["entry"][0]["resource"]["address"][0]["city"]
-        self.assertEqual(city_string, result_city_string)
+        for entry in response.data["results"]["entry"]:
```
spopelka-dsac: it's okay that addresses would be returned that don't contain the requested string, as long as at least one of the addresses for each practitioner does contain the requested string. This test would incorrectly fail for practitioners that have multiple addresses on different streets

backend/npdfhir/tests/test_practitioner.py
```diff
@@ -275,8 +288,9 @@ def test_list_filter_by_address_city(self):
...
```
spopelka-dsac: Same comment as above: It's okay if some of the cities are different as long as at least one city for each returned practitioner matches the filter. This would fail for a practitioner that has multiple addresses in different cities

backend/npdfhir/tests/fixtures/endpoint.py
IsaacMilarky: It was a while ago but I think I did that because in the reference data sql file it inserts the endpoint_connection_type data in the reference data but it doesn't insert anything for the endpoint_type data. I wanted to make sure that it would handle if that data was added in the reference data sql file.

backend/npdfhir/tests/fixtures/endpoint.py
```diff
@@ -0,0 +1,77 @@
...
+
+    if not ehr:
+        new_vendor_id = uuid.uuid4()
+        ehr_vendor = EhrVendor.objects.create(
+            id=new_vendor_id, name=f"My Sample{new_vendor_id}", is_cms_aligned_network=True
+        )
+    else:
+        ehr_vendor = ehr
+
+    et = EnvironmentType.objects.get(pk="prod")
```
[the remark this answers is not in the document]
spopelka-dsac: If you're passing the pk directly, you need to specify the field name as it appears in in the table (so the exact text in line 64 would be `environment_type_id="prod",`. It makes the code more streamlined

backend/npdfhir/tests/test_practitioner.py
```diff
@@ -245,28 +251,35 @@ def test_list_filter_by_npi_general(self):
         self.assertEqual(response.status_code, status.HTTP_200_OK)
         assert_has_results(self, response)
 
-        result_npi = response.data["results"]["entry"][0]["resource"]["identifier"][0]["value"]
-        self.assertEqual(self.pracs[0].npi.npi, int(result_npi))
+        for entry in response.data["results"]["entry"]:
```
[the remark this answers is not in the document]
spopelka-dsac: Okay, I think the test is asserting the correct things, but the way the assertion is structured is confusing and a bit misleading. 

Can you add a line at the start of this test that says `identifier = "1234567890"`
Then pass `identifier` in line 277 (instead of the hardcoded value) and line 283 (instead of `self.pracs[0].npi.npi`?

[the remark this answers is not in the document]
spopelka-dsac: But also we do want to make sure that both of them are being returned, so we can confirm that the identifier filter is operating across npi and other identifier... Even with my proposed change, it would be unclear whether both, one, or none are returned. Can you assert that both individual ids are present in the response?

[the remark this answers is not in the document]
IsaacMilarky: For sure I will implement that

PR #260, opened 2025-12-11
Backend test fix

IsaacMilarky:
## Problem

Test results publish action is not properly reporting status

## Solution

Remove the action at least for now


abachman-dsac (approved):
let's try this to get the release cut for #243 landed

PR #262, opened 2025-12-11
adding migration, updating models.py and updating serializers

spopelka-dsac:
## Problem

we had some incorrect data types based on data unknowns at the time

## Solution

* adds a migration to update the data types for a few fields
* updates models.py accordingly
* updates tests accordingly

## Test Plan

Note: a prior version of the sample data introduced a record with data that are not the same shape as the halloween data/ expected prod data, so you will need to delete that record locally in order for this migration to succeed locally.

```
delete from npd.organization_to_address where address_id = '55555555-5555-5555-5555-555555555555';
delete from npd.address where address_us_id like 'TEST00001';
delete from npd.address_us where id like 'TEST00001';
```

### DIFF
--- a/.github/workflows/backend-test.yml
+++ b/.github/workflows/backend-test.yml
@@ -6,9 +6,6 @@ permissions:
 
 on:
   pull_request:
-  push:
-    branches:
-      - main
 
 jobs:
   lint:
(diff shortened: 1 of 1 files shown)


abachman-dsac (approved):
i like the idea of a big negative `id` value showing up in a database 👍

PR #269, opened 2025-12-12
[NDH-471] Organization listing page with pagination

abachman-dsac:
[Jira Ticket NDH-471](https://jiraent.cms.gov/browse/NDH-471)

## Problem

Building on the work originally included in #257, we'd like to be able to navigate through multiple pages of results.

## Solution

Using [the CMS Design System Pagination component](https://design.cms.gov/components/pagination/?theme=core) and basic full page reloads for navigation, implement basic pagination.

I stuck the pagination controls at the top and bottom of the listing container to make them easier to find.

## Result

Pagination controls are visible on `/organizations` page when there are more than 10 records.

localhost_8000_organizations_paginated

Future work: 

- make `page_size` more dynamic
- load just the next page of data and rerender without a full browser reload

## Test Plan

Automated tests: TBD

Local dev: with existing development sample data and signed in as `developer@cms.hhs.gov`, visit http://localhost:8000/organizations, see the pagination controls and click around.

### DIFF
--- a/backend/npdfhir/management/commands/seedsystem.py
+++ b/backend/npdfhir/management/commands/seedsystem.py
@@ -1,8 +1,10 @@
 import json
+import random
 
 from django.core.management.base import BaseCommand
 from django.core.serializers.json import DjangoJSONEncoder
 from django.db import IntegrityError
+from faker import Faker
 
 from npdfhir.tests.fixtures import create_endpoint, create_organization, create_practitioner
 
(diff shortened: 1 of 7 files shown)


abachman-dsac:
> Dope! Looks great. Curiosity question... was it an intentional choice to populate a query string parameter with the page number and reload the whole page? It might be cool to store the page number in state and just reload the pagination component instead of the whole page.

The way I built it out here was the simplest first thing. 

Way I see it, going full browser defaults means it just works with the CMS Design System `<Pagination />` component , the URL in the browser always shows page number, and you can hot-link any page of results easily. 

Going React state plus browser navigation would've taken more work here.

PR #270, opened 2025-12-12
[NDH-278] Build out OrganizationAffiliation endpoint

IsaacMilarky:
## Problem

We currently do not have an endpoint that can query [OrganizationAffiliation](https://build.fhir.org/organizationaffiliation.html) data. This endpoint returns data that represents the affiliation relationship between clinical organizations and EHR Vendors

Additionally, we need to make it so that EHR Vendors can be queried in the Organization view. 

## Solution

Add a new view set for OrganizationAffiliation data that has list and retrieve endpoints. Also, create a corresponding serializer class that will create FHIR JSON for the data returned. 

Additionally, make it so that EHR Vendors can be queried in the Organization viewset. 

Tests also need to be created for all this.  

## Result

Summary:

- Register new OrganizationAffiliation viewset 
- Add new class OrganizationAffiliationSerializer
- Add new OrganizationAffiliation viewset class and create query of the Organization table with all the neccessary data attached or annotated with subqueries
- Add new helper class FHIROrganizationSource that acts as an interface so that EHRVendor objects can be treated as Organizations for the purpose of serialization
- Edit the Organization viewset to also query EHRVendors 
- Alter the Organization viewset to apply filtersets manually separately to orgs and ehrvendor objects
- Add an ehr_vendor_filterset file to apply the same filters that orgs can have to EHRVendors
- Add each retrieved record as a list of FHIROrganizationSource objects and then sort based on name by default or the param processed by ParamOrderingFilter
- Edit the OrganizationSerializer to also support EHRVendor data. 

## Test Plan

`make test`


spopelka-dsac (changes requested):
It will be easier to assess the rest of the changes once the test fixtures PR is landed, so that this diff isn't so big. I'll provide further comments once that one is merged

backend/npdfhir/serializers.py
```diff
@@ -434,6 +458,153 @@ def to_representation(self, instance):
...
+
+        organization_affiliation.participatingOrganization = genReference("fhir-organization-detail", instance.id, request)
+        organization_affiliation.participatingOrganization.display = str(instance.organization_name)
+
+        # NOTE: Period for OrganizationAffiliation cannot currently be fetched so its blank
+
+        organization_affiliation.network = [genReference("fhir-organization-detail", instance.id, request)]
+        organization_affiliation.network[0].display = str(instance.organization_name)
+
+        organization_affiliation.code = codes
```
spopelka-dsac: Per the FHIR spec, "code" here refers to the [Organization Role codeset](https://hl7.org/fhir/R4B/valueset-organization-role.html). Since the relationships we are modeling here with OrganizationAffiliation are EHR Vendor to organization that uses that EHR vendor, "HIE/HIO | HIE/HIO | An organization that facilitates electronic clinical data exchange between entities" is the most appropriate code to populate here.

backend/npdfhir/views.py
```diff
@@ -11,20 +12,23 @@
 from rest_framework.views import APIView
 from rest_framework.renderers import BrowsableAPIRenderer
 from rest_framework.response import Response
-from rest_framework.filters import OrderingFilter
+from rest_framework.filters import OrderingFilter, SearchFilter
```
spopelka-dsac: SearchFilter should be removed

spopelka-dsac (changes requested):
Can you please fully move filters and sort to a separate PR? It will make easier to review filter and sort in future PRs when we get to that.

Can you please also delete the commented code in the OrganizationAffiliaition serializer? At this point, I don't anticipate that we will be adding those elements in.

PR #271, opened 2025-12-15
EC2 Instance user data for semi-ephemeral EC2 GitHub Actions Runners

wbprice:
## Problem

- Post-install configuration of EC2 instances for GitHub Actions is pretty tedious
- I wanted to use Docker containers in Fargate to build docker image but discovered that won't work
    - kaniko is deprecated, not a drop-in docker replacement
    - buildah is kind of a docker replacement but building images requires permissions that we can't grant in a fargate environment
- This PR makes the experience of setting up EC2 instances for GHA much less manual

## Solution

Creates a bootstrapping script, passed to the EC2 instance user_data property, that does a lot of the setup work currently in README.md.

## Result

In dev, a second GitHub Actions runner is available for CI workloads.

## Test Plan

Merge it


spopelka-dsac (approved):
🙌

PR #273, opened 2025-12-15
[NDH-539] Add Parameters Missing in FHIRPractitionerRoleViewSet

IsaacMilarky:
## Problem

We are aiming to have a set of search entry points/filters in the FHIR api views. We do not currently have all of the ones outlined in the FHIR API Endpoint Search document. 

## Solution

Add many of the missing searches and filters to the practitioner role filter set file. 

## Result

Summary:

* Add filter to filter the organization type
* Add filter to filter between active and non-active roles
* Add filter to filter by practitioner identifier 
* Add filter to filter by Nucc/snomed specialty 
* Add endpoint filters to filter by endpoint connection, payload type, org id, and org name
* Add location filters for address, city, zip code, state
* Add tests for all of these
* Still missing PractitionerRole searches for endpoint status since that is not yet implemented
* Still missing Location distance from a point which I can do in a separate PR

## Test Plan

Run `make test`

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -8,10 +8,8 @@
 # The Repo Domains section of your CODEOWNERS.md file helps manage code review responsibilities efficiently. Each domain represents a different aspect of the repository, such as documentation, frontend, backend, DevOps, testing, etc. In this section, list each domain and assign the appropriate GitHub usernames or teams responsible for that domain. This ensures that pull requests (PRs) are reviewed by the right experts, maintaining high code quality and relevance
 # Not every name on the list is required to review every change, but at least one person named should be review every PR making contributions.
 
-/frontend/ @sachin-panayil @sman-dsac @spopelka-dsac
+/frontend/ @sachin-panayil @sman-dsac @spopelka-dsac @wbprice
 
-/backend/ @spopelka-dsac @IsaacMilarky @sachin-panayil
+/backend/ @spopelka-dsac @IsaacMilarky @sachin-panayil @wbprice
 
-/infrastructure/ @wbprice
-
-/flyway/ @spopelka-dsac
+/flyway/ @spopelka-dsac @wbprice


backend/npdfhir/filters/practitioner_role_filter_set.py
```diff
@@ -52,11 +134,87 @@ def filter_practitioner_gender(self, queryset, name, value):
...
+    def filter_endpoint_organization_id(self, queryset, name, value):
+        #The parent of the organization that owns the location the endpoint is attached to
+        return queryset.filter(
+            location__organization__parent__id=value
+        )
+    
+    def filter_endpoint_organization_name(self, queryset, name, value):
+        #The parent of the organization that owns the location the endpoint is attached to
+        return queryset.filter(
+            location__organization__parent__organizationtoname__name=value
```
spopelka-dsac: This was a lack of clarity on my part in the search doc; This would be the organization_name, not the parent organization_name

backend/npdfhir/filters/practitioner_role_filter_set.py
```diff
@@ -25,13 +26,94 @@ class PractitionerRoleFilterSet(filters.FilterSet):
...
+
+    location_zip_code = filters.CharFilter(
+        method="filter_address_postalcode",
+        help_text="Filter by the location postal code"
+    )
+
+
     class Meta:
         model = ProviderToLocation
         fields = [
```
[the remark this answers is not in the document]
IsaacMilarky: This is missing in the current schema. I will add this field to the EndpointInstance table I was confused if this was in the scope of this ticket before.

flyway/sql/migrations/V16__add_endpoint_status_column.sql
```diff
@@ -0,0 +1 @@
+alter table npd.endpoint_instance add endpoint_status varchar(64);
```
spopelka-dsac: Sorry, this is so nitpicky, but by our sql naming conventions this should just be "status" not "endpoint_status"

spopelka-dsac (changes requested):

PR #274, opened 2025-12-16
update weekly refresh token using date modulo'd by 7 rather than mill…

wbprice:
Update weekly refresh token to use days rather than milliseconds (Terraform doesn't support this but ChatGPT sure thinks it does).


abachman-dsac (approved):
ChatGPT said this is fine 🤖 👍

PR #275, opened 2025-12-16
Update trust policy to read from npd-ops rather than NPD

wbprice:
## Problem

We want to change the name of the npd repo to npd-ops to better describe what that repo is for

## Solution

Update the trust policy for the github runner

## Result

This needs to run first (to update the policy), then the repo CMS-Enterprise/npd needs to be renamed to CMS-Enteprise/npd-ops. As the admin for CMS-Enterprise/npd-ops, I can change this myself

## Test Plan

Merge it


spopelka-dsac (approved):

abachman-dsac (approved):
reduce confusion! 

i'm for it

PR #281, opened 2025-12-17
Wbprice/ndh 584 resolve flyway schema already exists error

wbprice:
## Problem

flyway-error

When running flyway migrations against our databases, they complete successfully the first time, but then fail with the attached error message. Even though flyway has created the schema_history_table and populated it with the schema history once, it can't seem to find that table again and thinks that the database has already been set up and is being managed external to flyway.

I see this as a consequence of a mismatch between where the flyway_schema_history table is created and where flyway expects it to be. @abachman and I have fixed this manually in the past by moving the flyway_schema_history table from public to the npd schema.

## Solution

- Tell Flyway to put the flyway_schema_history table on the `public` schema (where it expects it to exist by default)

## Result

Migration command works consistently

(on a test new schema)

Screenshot 2025-12-17 at 4 03 37 PM

## Test Plan

Merge it, see if the error persists for a new database instance / schema in dev.


tspecht-cms (approved):

flyway/sql/migrations/V1__initial_npd_schema.sql
```diff
@@ -21,7 +21,7 @@ SET row_security = off;
 -- Name: npd; Type: SCHEMA; Schema: -; Owner: -
 --
 
-CREATE SCHEMA npd;
```
abachman-dsac: note: this change will cause this `.sql` file to be picked up by flyway as modified and will require `repair` to be run on any system which has already applied it, including all our deployments. 

e.g., when pulling this branch to my machine and running it against a database which was previously migrated from `main`, i see: 
 
```
$ docker compose run --rm db-migrations migrate
[+] Creating 1/1
 ✔ Container npd-db-1  Recreated                                                                           0.2s
[+] Running 1/1
 ✔ Container npd-db-1  Started                                                                             0.1s
Flyway OSS Edition 10.22.0 by Redgate

See release notes here: https://rd.gt/416ObMi
Database: jdbc:postgresql://db:5432/npd_development (PostgreSQL 17.6)

ERROR: Validate failed: Migrations have failed validation
Migration checksum mismatch for migration version 1
-> Applied to database : 221568673
-> Resolved locally    : -869299429
Either revert the changes to the migration, or run repair to update the schema history.
Need more flexibility with validation rules? Learn more: https://rd.gt/3AbJUZE
```

If it is included, the PR description should have a plan for running flyway's `repair` command for every existing instance of the npd database.

abachman-dsac (changes requested):

infrastructure/modules/fhir-api/main.tf
```diff
@@ -149,7 +149,11 @@ resource "aws_ecs_task_definition" "app" {
       environment = [
         {
           name  = "FLYWAY_URL"
-          value = "jdbc:postgresql://${var.db.db_instance_address}:${var.db.db_instance_port}/${var.db.db_instance_name}?options=--search_path%3Dnpd%2Cpublic"
+          value = "jdbc:postgresql://${var.db.db_instance_address}:${var.db.db_instance_port}/${var.db.db_instance_name}"
+        },
+        {
+          name  = "FLYWAY_DEFAULT_SCHEMA",
+          value = "npd"
```
spopelka-dsac: When I ran flyway with `FLYWAY_DEFAULT_SCHEMA = npd` against my local database (which has is current through v14 and has the flyway schema history tracked in `public.FLYWAY_SCHEMA_HISTORY`), I received the following error:

Image

I tested it with `FLYWAY_DEFAULT_SCHEMA = public` and it was able to run without error. Would you be okay with setting this value to public here and then not modifying the V1 migration? I think that is the easiest/ cleanest path forward

spopelka-dsac (approved):
Thanks!

PR #282, opened 2025-12-17
[NDH-639] Add optional SQL query tracing to the backend test suite

abachman-dsac:
[Jira Ticket NDH-639](https://jiraent.cms.gov/browse/NDH-639)

## Problem

We have the django-debug-toolbar for viewing SQL queries, one in-browser HTTP request at a time, but nothing in the test suite to do the same.
 
## Solution

Using a combination of functionality built into Django and structlog, this PR introduces (optional) structured SQL query logging from the backend test suite.

## Result / Usage

Setting the `SQL_TRACING` environment variable to `True` will cause the backend test suite to log all SQL queries to `backend/django_queries.log` in JSON format:

```json
{
  "event": "sql_query",
  "ip": "127.0.0.1",
  "test": "npdfhir.tests.test_organization.OrganizationViewSetTestCase.test_list_default",
  "request_id": "da4d5603-1148-4ec4-ad23-96ebc4a7b0a6",
  "timestamp": "2025-12-17T22:04:48.897900Z",
  "logger": "django.db.backends",
  "level": "debug",
  "alias": "default",
  "sql": "SELECT \"clinical_organization\".\"organization_id\", \"clinical_organization\".\"npi\" FROM \"clinical_organization\" WHERE \"clinical_organization\".\"organization_id\" IN ('c591bfc5b4ed49af926f569056b5b1aa'::uuid, '5f56f3f03bd642ceb275f12f92a4ba40'::uuid, '682af722514f474f96637cfb81ce5f40'::uuid, 'b76b81a5bc6c4f2787dd2c76aa3fdcc7'::uuid, '256a67271c17409197bb9528f0a20a13'::uuid, '673716bfb0b548918b5f9eaba0dfa841'::uuid, 'ba44e463bee549a2a01802eb344dbaa2'::uuid, '2afca474a476443ba8924f254ac01096'::uuid, '64407cecaa4149a6a7955969a9155eb5'::uuid, '4a2780f8a00b44939255d995e2dfc383'::uuid)",
  "args": [],
  "duration": 0.00017787500109989196,
  "created": "2025-12-17T22:04:48.897871"
}
```

The fields unique to this being a sql query trace are: 

- `event`: always has the value `"sql_query"`
- `sql`: the actual query
- `args`: bound arguments if the query uses placeholders
- `alias`: [a custom name for this query](https://docs.djangoproject.com/en/6.0/ref/models/querysets/#alias), defaults to`"default"`
- `duration`: the duration of the query in seconds
- `created`: an `isoformat` datetime tracking when the query was executed
- `test`: the test which caused the query to execute

You can set the env var and run the test suite with: 

```sh
bin/npr -t -e SQL_TRACING=True python manage.py test 
```

or try it on a single test with: 

```sh
bin/npr -t -e SQL_TRACING=True python manage.py test npdfhir.tests.test_organization.OrganizationViewSetTestCase.test_list_default
```

### DIFF
--- a/backend/app/settings.py
+++ b/backend/app/settings.py
@@ -311,12 +311,6 @@ LOGGING = {
             "class": "logging.StreamHandler",
             "formatter": "json_formatter",
         },
-        "queries_file": {
-            "level": LOG_LEVEL,
-            "class": "logging.FileHandler",
-            "filename": "django_queries.log",
-            "formatter": "json_formatter",
-        },
     },
     "loggers": {
         "": {
@@ -337,6 +331,12 @@ if SQL_TRACING:
     )
 
     # write SQL query logs to backend/django_queries.log
+    LOGGING["handlers"]["queries_file"] = {
+        "level": LOG_LEVEL,
+        "class": "logging.FileHandler",
+        "filename": "django_queries.log",
+        "formatter": "json_formatter",
+    }
     LOGGING["loggers"]["django.db.backends"] = {
         "handlers": ["queries_file"],
         "level": "DEBUG",


spopelka-dsac (approved):
very cool! Great addition

PR #459, opened 2026-04-03
Unified Search Experience

sachin-panayil:
## Problem

The existing search experience required users to know in advance whether they were looking for a
practitioner or an organization, routing them through two separate search pages
(/practitioners/search and /organizations/search). There was no way to search across both resource
types at once, or to find a practitioner by the organization they work at.

## Solution

Added a unified search experience at /search that replaces the old search hub landing page. The new
page provides a single form with four fields — provider name, organization name, NPI, and location —
and intelligently routes the query based on what's filled in:

- Provider name only → searches practitioners
- Organization name only → searches organizations
- NPI only → looks up either resource type by NPI
- Provider name + organization name → cross-entity search via PractitionerRole, showing practitioners filtered by their affiliated organization
- Location → refines any of the above; location alone is not a valid search and shows a hint

## Result

Users can now search for both practitioners and organizations from a single entry point at /search.
Cross-entity searches (provider at organization) are supported without requiring two separate
queries. The search button is disabled unless at least one of provider name, organization name, or
NPI is filled, preventing empty searches.

## Test Plan
- run `make up`
- go through search experience

### DIFF
--- a/frontend/src/@types/unifiedSearch.d.ts
+++ b/frontend/src/@types/unifiedSearch.d.ts
@@ -1,0 +1,14 @@
+interface UnifiedSearchParams {
+  providerName?: string
+  npi?: string
+  location?: string
+  organizationName?: string
+}
+
+interface UnifiedSearchURLParams extends PaginationParams {
+  provider_name?: string
+  npi?: string
+  location?: string
+  organization_name?: string
+  sort?: string
+}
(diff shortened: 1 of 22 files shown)


spopelka-dsac (approved):
Approving this without review, due a quick turnaround on the leadership decision to do a front end overhaul. Hopefully we will have an opportunity to improve this code and confirm its soundness/validity after 4/9.

PR #469, opened 2026-04-08
[NDH-1044] About page content updates

sman-dsac:
Update the about page with updated content described in [NDH-1044](https://jiraent.cms.gov/browse/NDH-1044). Note that banner and nav updates are not implemented in this story, but the global updates are described elsewhere.


spopelka-dsac (approved):
lgtm, thanks!

spopelka-dsac:
@sman-dsac I can update the front end tests

PR #479, opened 2026-04-24
DB: add provider credential description to provider_to_credential

rmillergv:
## Summary

Adds `credential_description` to `npd.provider_to_credential`.

## What Changed

- adds Flyway migration `V22__add_provider_to_credential_description.sql`
- updates `backend/npdfhir/models.py` to include the new `credential_description` field on `ProviderToCredential`

## Why

This is a non-breaking schema extension that makes a credential description field available in the API-serving database for future use.

The paired `npd_etl` change will populate this field from PECOS-sourced credential description data during the Stage 3 load.

## Behavior

- no existing field is removed or changed
- no existing API/Search UI behavior changes as part of this PR
- current API/Search UI code would have to explicitly choose to read and use this field before it becomes user-visible

## Local Validation

- ran Flyway locally against the laptop test database (`npd_test`)
- confirmed the migration applied successfully through `V22`
- verified the new column exists in local Postgres:
  - `npd.provider_to_credential.credential_description`
- ran `python manage.py check` in the local test stack
- Django system check passed with no issues

## Notes

This PR only makes the PECOS-sourced credentials description available for future use.


rmillergv:
messed up, files having nothing to do with my change got pulled in

PR #483, opened 2026-05-11
Update silo static site to read manifest json

wbprice:
image

## Problem

The static site expects Palantir data exports files to be renamed before they can be published by the website

## Solution

Update the backend processes in app.py to read the manifest.json to fetch accurate file names rather than expecting the names to be consistent between releases.

## Result

- It should be easier to push updates to the static site when a data release is created
- Original size, compressed size, compression ratio values populate as expected

## Test Plan

```
kion f # assume prod role
docker compose up --build
```

1. Open `localhost:8080`
2. Click download buttons of various resources
3. Click load sample buttons of various resources

### DIFF
--- a/silo_site/Dockerfile
+++ b/silo_site/Dockerfile
@@ -10,4 +10,6 @@ COPY app.py .
 ADD static ./static
 # Run the app on 8080
 EXPOSE 8080/tcp
+ENV PYTHONUNBUFFERED=1
+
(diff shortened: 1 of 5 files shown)


pfiscina-doge (commented):
this is fine, thanks

rmillergv (approved):

